PR #9, opened 2018-05-03
Change TJPCosmo note format

elisachisari:
Make the format of the note agree with the templade for DESC pipelines. To compile the tex, go to `doc/tjpcosmo_note` and run `make lsstdescnote`.


elisachisari:
@joezuntz I don't know if we still want to have a note, but compilation now works.

elisachisari:
I've been giving the idea of the note a bit more thought. I think we should maintain a minimal note where we describe any systematics implementation that is inherent to firecrown and not described in CCL. No installation instructions, no API, since all of this is in the readthedocs page. Thoughts? @elikrause @beckermr @joezuntz @empEvil

beckermr:
Go for it!

elikrause:
Yes, this content seems a useful starting point for new users/developers! Should this be an actual note (pdf), on readthedocs, or something else entirely?

beckermr:
We could add notes to the module doc strings and just go with readthedocs.

vitenti:
This PR discuss some possible documentation styles that could be included in this repository:

1. API documentation in doc-strings
2. Physics description of likelihood implementations
3. Installation and beginning user documentation
4. Beginning developer documentation

The first one is already being set in place during the refactoring. The next issues 2, 3 and 4 will be discussed in three different issues #155, #153 and #154 respectively.

PR #34, opened 2018-10-18
ENH added version

### LINKED ISSUE #24: add `_version.py` file and integrate into `setup.py`
We need a `_version.py` file with the current version `0.1.0` in it. We also need the `setup.py` to read the version. 

Adding these lines to the `setup.py`

```python
__version__ = None
pth = os.path.join(os.path.dirname(os.path.realpath(__file__)), "firecrown", "_version.py")
with open(pth, 'r') as fp:
    exec(fp.read())
```

and putting the `__version__` string in `_version.py` should do the trick. Finally, we need to specify the version in the call to setup via `version=__version__`. Mad props for importing the version to the `firecrown/__init__.py` as well!

beckermr:
This PR adds the version.

Closes #24


beckermr:
Ack. Good catch! I forgot. 🤦‍♂️ 

It should be good to go now @joezuntz !

PR #35, opened 2018-10-18
ENH run metadata

### LINKED ISSUE #32: compute a unique hash and other metadata for each call of firecrown
Every call of firecrown needs to have metadata associated with it so we can track what was done.

This should include:

1. a copy of the config file
2. a unique hash associated with each run of the code (called the analysis id)
3. The current version of the code. 
4. The current version of CCL.
5. The git hash of the code if available. 

This data should be written in YAML to an output file, in say

```bash
output_<analysis_id>/metadata.yaml
output_<analysis_id>/config.yaml
```

beckermr:
This PR adds code for writing run metadata.

Closes #32

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,6 +56,7 @@ jobs:
           command: |
             . venv/bin/activate
             flake8 firecrown
+            flake8 bin/firecrown
 
       - run:
           name: test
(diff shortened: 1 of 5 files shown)


beckermr:
This should be merged after #34

beckermr:
@joezuntz this PR is good to go

joezuntz:
This is fine for now, but we should pass this metadata on to the cosmosis interface to be output as metadata too. I can do that later on.

beckermr:
Sure go for it.

PR #36, opened 2018-10-23
ENH added i/o stuff

### LINKED ISSUE #31: finish the firecrown statistic io
Right now each log-likelihood section returns any data as a set of possibly nest dictionaries containing values which are numpy structured arrays. We should add some temporary code to render this into a set of directories (following the nesting of the dictionaries) with CSV outputs for each record array. 

To be more specific, something like

```python
{'a': {'carr': <structured array>}, 'darr': <structured array>}
```

should end up as 

```bash
output_<analysis_id>/a/carr.csv
output_<analysis_id>/darr.csv
```

This will have to be revisited, but should work for now.

beckermr:
This PR adds i/o.

Closes #31

### DIFF
--- a/bin/firecrown
+++ b/bin/firecrown
@@ -46,11 +46,13 @@ def run(action, config, output):
 
     elif action == 'run-cosmosis':
         firecrown.run_cosmosis(_config, data)
+        stats = None
 
     # we associate a unique id with each analysis
     # let's write that out with some extra info
     write_metadata(analysis_id, output, config)
-    write_statistics(analysis_id, output, stats)
+    if stats is not None:
+        write_statistics(analysis_id, output, stats)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


bin/firecrown
```diff
@@ -49,6 +50,7 @@ def run(action, config, output):
     # we associate a unique id with each analysis
     # let's write that out with some extra info
     write_metadata(analysis_id, output, config)
+    write_statistics(analysis_id, output, stats)
```
joezuntz: stats is currently undefined here if we use the run-cosmosis branch

beckermr: Ack. Good catch and thank you! Fixed!

beckermr:
@joezuntz changed. Anything else?

beckermr:
bump here - I am cool merging without an official approval too 🤷‍♂️

beckermr:
@joezuntz Elisabeth is OK merging here, so I am going to go ahead. Happy to revisit in future PRs if you have further comments!

PR #58, opened 2019-01-25
Magnification bias

### LINKED ISSUE #51: Magnification bias
Implement magnification bias model for  galaxy number contrast following Joachimi and Bridle (2010),  arXiv:0911.2454. Incorporate into correlation functions and marginalize over uncertainty in the galaxy number count luminosity slope.

zdu863:
Implement magnification bias model (resolves #51).

### DIFF
--- a/firecrown/ccl/systematics/lss.py
+++ b/firecrown/ccl/systematics/lss.py
@@ -3,7 +3,7 @@ import numpy as np
 
 from ..core import Systematic
 
-__all__ = ['LinearBiasSystematic', 'MagnificationBias']
+__all__ = ['LinearBiasSystematic', 'MagnificationBiasSystematic']
 
 
 class LinearBiasSystematic(Systematic):
@@ -49,7 +49,7 @@ class LinearBiasSystematic(Systematic):
         source.bias_ *= pref
 
 
-class MagnificationBias(Systematic):
+class MagnificationBiasSystematic(Systematic):
     """Magnification bias systematic.
 
     This systematic adds a magnification bias model for galaxy number contrast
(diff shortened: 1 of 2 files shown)


firecrown/ccl/systematics/lss.py
```diff
@@ -1,8 +1,9 @@
 import pyccl as ccl
+import numpy as np
 
 from ..core import Systematic
 
-__all__ = ['LinearBiasSystematic']
+__all__ = ['LinearBiasSystematic', 'MagnificationBias']
```
beckermr: We should stick with the naming convention by adding `Systematic` to the end. Or we should remove this from the linear bias. Either way they should be the same.

firecrown/ccl/systematics/lss.py
```diff
@@ -46,3 +47,49 @@ def apply(self, cosmo, params, source):
...
+    This systematic adds a magnification bias model for galaxy number contrast
+    following Joachimi & Bridle (2010), arXiv:0911.2454.
+
+    Parameters
+    ----------
+    r_lim : str
+        The name of the limiting magnitude in r band filter.
+    Sig_c, eta, z_c, z_m : str
+        The name of the fitting parameters in Joachimi & Bridle (2010) equation
+    (C.1).
```
beckermr: Indent this four spaces.

firecrown/ccl/systematics/lss.py
```diff
@@ -46,3 +47,49 @@ def apply(self, cosmo, params, source):
...
+
+    Methods
+    -------
+    apply : appaly the systematic to a source
+    """
+    def __init__(self, r_lim, Sig_c, eta, z_c, z_m):
+        self.r_lim = r_lim
+        self.Sig_c = Sig_c
```
beckermr: Generally all python parameters should be lowercase.

firecrown/ccl/systematics/lss.py
```diff
@@ -46,3 +47,49 @@ def apply(self, cosmo, params, source):
...
+        source : a source object
+            The source to which apply the shear bias.
+        """
+
+        z_bar = params[self.z_c] + params[self.z_m] * (params[self.r_lim] - 24)
+        z = source.z_
+        s = (
+            params[self.eta] / params[self.r_lim] - 3 * params[self.z_m] /
+            z_bar + 1.5 * params[self.z_m] * np.power(z / z_bar, 1.5) / z_bar)
+        source.mag_bias_ = s / np.log(10)
```
beckermr: We should be replacing this attribute. Instead we should be using something like `*=` here. This will let people declare an overall amplitude that can be a parameter as well.

PR #64, opened 2019-04-15
Photozsystematics

zdu863:
Based on this [script](https://gist.github.com/MarkusMichaelRau/8ee36f06a34a74abda7b3853be2b60ba), the PR adds a systematic through convolving a redshift distribution with a corresponding error distribution according to [this paper](https://arxiv.org/pdf/1002.3383.pdf) Eq. 5.

### DIFF
--- a/firecrown/ccl/systematics/pz.py
+++ b/firecrown/ccl/systematics/pz.py
@@ -52,8 +52,15 @@ class PhotoZSystematic(Systematic):
 
     Parameters:
     -----------
-    mu_0, mu_1: parameters that describes the evolution of the mean function
-    sigma: parameter that describes the evolution of the std function
+    mu_0 : str
+        the name of the first parameter that describes the evolution of the
+        mean function
+    mu_1 : str
+        the name of the second parameter that describes the evolution of the
+        mean function
+    sigma : str
+        the name of the parameter that describes the evolution of the
+        std function
 
     Methods
     -------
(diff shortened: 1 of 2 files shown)


firecrown/ccl/systematics/pz.py
```diff
@@ -37,3 +39,54 @@ def apply(self, cosmo, params, source):
...
+        ----------
+        cosmo : pyccl.Cosmology
+            A pyccl.Cosmology object.
+        params : dict
+            A dictionary mapping parameter names to their current values.
+        source : a source object
+            The source to which apply the shift.
+        """
+        zorig, N = (source.z_, len(source.z_))
+        zpdf_orig = source.dndz_/np.sum(source.dndz_*source.z_)
```
beckermr: This confuses me. z_ is the array of z values, not the bin spacing. Also, dndz is already normalized here.

firecrown/ccl/systematics/pz.py
```diff
@@ -37,3 +39,54 @@ def apply(self, cosmo, params, source):
...
+        joint_distr = np.zeros((N, N))
```
beckermr: There is no reason to keep the full array of values around.

firecrown/ccl/systematics/tests/test_pz.py
```diff
@@ -58,3 +59,29 @@ def test_photoz_shift_fill_nan():
     msk = z - 0.5 <= 0
     assert np.all(src.dndz_[msk] == 0)
     assert np.all(src.dndz_[~msk] != 0)
+
+
+def test_photoz_systematic_smoke():
```
beckermr: This test doesn't actually test anything. You are setting an attribute and then comparing it to what you set it to.

vitenti:
This PR contain code for applying a Gaussian distribution to a dndz. We are planning to have a separate object to describe dndz and this code may be used there in the future. I'm closing this PR and any useful code will be copied to the new structure.

PR #78, opened 2020-01-31
more flexible config

slosar:
This PR implements three small changes to parers:
i) instead of filename, the first parameter is source. If source has a read() method, it will use it to get yaml, otherwise it is interpreted as filename
ii) You can supply an optional "memory_data", that is a dictionary that gets pasted on top of whatever was parsed from yaml. E.g. you can do constructions like:
```       config, data = firecrown.parse(io.StringIO(fc_config),
	       memory_data={'two_point':{'sacc_data':sacc},'parameters':fid_params})
```
iii) For `ccl.twpoint` you can supply either `sacc_file` as filename or `sacc_data`, which is a sacc object. Together with ii) it allows you to pass an in-memory sacc object to firecrown gracefully.


beckermr:
`flake8` is going to hate you. You'll need to clean up the code a bit. You can run it locally. I run it all the time in my editor.

firecrown/parser.py
```diff
@@ -1,30 +1,48 @@
...
         The raw config file as a dictionary.
     data : dict
         A dictionary containg each analyses key replaced with its
         corresponding data and function to compute the log-likelihood.
     """
 
-    with open(filename, 'r') as fp:
-        config_str = jinja2.Template(fp.read()).render()
+    if  not hasattr(source,"read"):
+        source = open (source,'r')
```
beckermr: You are leaking an open file handle here.

firecrown/parser.py
```diff
@@ -1,30 +1,48 @@
...
     """Parse a configuration file.
 
     Parameters
     ----------
-    filename : str
+    source : str
         The config file to parse. Should be YAML formatted.
+        if source has 'read' attribute, it is used to fetch data,
+        otherwise it is interpreted as a filename
+    memory_data : (optional). Dictionary (can be of dictionaries)
```
beckermr: This parameter name is not great and the docstring is not formatted properly.

firecrown/parser.py
```diff
@@ -1,30 +1,48 @@
...
 
-    with open(filename, 'r') as fp:
-        config_str = jinja2.Template(fp.read()).render()
+    if  not hasattr(source,"read"):
+        source = open (source,'r')
+    config_str = jinja2.Template(source.read()).render()
     config = yaml.load(config_str, Loader=yaml.Loader)
     data = yaml.load(config_str, Loader=yaml.Loader)
 
+    ## if stuff came already parsed, bring it in
```
beckermr: please use only one `#` for comments

firecrown/parser.py
```diff
@@ -1,30 +1,48 @@
...
+        ## recursive update of potentially nested dictionaries
+        def update_dict(d, u):
+            for k, v in u.items():
+                if isinstance(v, collections.abc.Mapping):
+                    d[k] = update_dict(d.get(k, {}), v)
+                else:
+                    d[k] = v
+            return d
+        data = update_dict (data, memory_data)
+        config = update_dict (config, memory_data)
```
beckermr: I need to think through this change and what it might mean. Letting people make essentially arbitrary updates to the config structure and the data could easily break things elsewhere in unexpected ways.

firecrown/ccl/two_point.py
```diff
@@ -28,10 +28,14 @@ def parse_config(analysis):
     if 'likelihood' in analysis:
         new_keys['likelihood'] = _parse_likelihood(analysis['likelihood'])
 
-    # read data if there is a sacc file
-    if 'sacc_file' in analysis:
+    # read data if there is a sacc file or read object
+    sacc_data = None
+    if 'sacc_data' in analysis:
+        sacc_data = analysis['sacc_data']
+    elif 'sacc_file' in analysis :
```
beckermr: We should use a single keyword here called `sacc_data` that is either the object or a string that gets read into a file. There is no reason to have two and that will make things simpler.

slosar: Fixed that. Kept sacc_file with deprecation warning.

slosar:
> OK. So right now, firecrown always writes back out the config you used to the final analysis. This way you can always reproduce things. The changes here potentially break that convention in the code. I'd say for in-memory use, we should simply allow the passed in config to be a dictionary and put in on the user to combine things etc.

I think that for in-memory use we do not want the user to be mucking with config after we have parsed it, because we want to catch special cases, but also because it looks very hackish. We could allow user to input a general dictionary (where values can be anything), but I do like yaml parsing, it does make things more modular. I.e you could imagine all but a few keys living in an actual yaml file and you only change in-memory ones that you need.

firecrown/ccl/two_point.py
```diff
@@ -28,10 +29,18 @@ def parse_config(analysis):
...
         new_keys['likelihood'] = _parse_likelihood(analysis['likelihood'])
 
-    # read data if there is a sacc file
+    # read data if there is a sacc file or read object
     if 'sacc_file' in analysis:
-        sacc_data = sacc.Sacc.load_fits(
-            os.path.expanduser(os.path.expandvars(analysis['sacc_file'])))
+        warnings.warn("sacc_file is deprecated, use sacc_data instead",
+                      DeprecationWarning)
+        analysis['sacc_data'] = analysis['sacc_file']
```
beckermr: This warning is nice but not needed. I would rather us make the hard change now (i.e. no longer support this) and go through and adjust all of the configs we have. I don't want to carry this around forever.

firecrown/ccl/two_point.py
```diff
@@ -28,10 +29,18 @@ def parse_config(analysis):
...
+    if 'sacc_data' in analysis:
+        if type('sacc_data') == str:
```
beckermr: the syntax here is wrong. type('str') is always str. You need to grab the key in the analysis dict. Also please use `isinstance(s, str)`. This will catch subclasses and inheritance.

firecrown/parser.py
```diff
@@ -1,30 +1,56 @@
...
+        The raw config file as a dictionary, updated with settings
     data : dict
         A dictionary containg each analyses key replaced with its
         corresponding data and function to compute the log-likelihood.
     """
 
-    with open(filename, 'r') as fp:
-        config_str = jinja2.Template(fp.read()).render()
+    if hasattr(source, "read"):
+        yamlconf = source.read()
```
beckermr: This change won't work. The parser passes in a filename. A simple if that reads with a context manager if needed will be fine.

beckermr: ack this comment is wrong sorry!

slosar:
> I don't think it is a good idea to let people override both the config and data dictionaries using the same dictionary (or at all). These are different things.
>
> Can you send an example piece of code illustrating what you want to do? This might help us find a better solution.

See link 296 of https://github.com/LSSTDESC/TXPipe/blob/blinding/txpipe/twopoint.py
```
        config, data = firecrown.parse(io.StringIO(fc_config),
               settings={'two_point':{'sacc_data':sacc},'parameters':fid_params})
        fidcosmo = firecrown.get_ccl_cosmology(config['parameters'])
        _, fidstat = firecrown.compute_loglike(cosmo=fidcosmo, data=data)
```

I need to change config, because I'm passing cosmological parameters, which I need to right for `get_ccl_cosmology`. I also need to change data becuse I'm passing sacc file as file object. 
My objective here is to:
 * configure via yaml
 * have ability to pass a sacc object without ever having to dump it to disk
 * pass parameters without turning them into string only for yaml parser to turn them back to float (although I am willing to give up on this)

slosar:
(also I think I'm failing test, because `sacc_file` option is now gone)

PR #89, opened 2020-02-25
ENH cluster sources

beckermr:
This PR contains the cluster source API based on the new halo model functionality put into CCL. I am starting over here to try and get a solid set of core things put down in code before a lot of the other items are added.


beckermr:
OK @slosar @elikrause @damonge - I have taken a pass at the API for clusters here. This work builds on stuff @elikrause was working on, but I started over here in a new PR to clean things up. 

A few things

1. I introduce two key concepts for clusters, a MOR and a selection function.
2. The selection function is always expressed as a function of mass and scale factor. This means that given some selection in lambda, the code will have to do some sort of integral to get the final selection function.
3. Each cluster source gets a default top-hat selection function and identity mass observable relationship.

My question (mostly for @damonge) is how much of this should be moved into the expanding halo model capabilities of CCL? 

Right now I am at the point where I need to do two 1-pt halo model integrals over the mass and bias functions. CCL already has parts of these integrals lying around I think, but it is not 100% clear to me (even after doing CR on those PRs).

beckermr:
@damonge I think I could add an appropriate method here

https://github.com/LSSTDESC/CCL/blob/master/pyccl/halos/halo_model.py#L140

that uses the integration routines that already exist?

damonge:
@beckermr sorry for the delay.
I wouldn't pollute CCL with survey-specific things, but I think what you say makes sense. It would be useful to have a method within HMCalculator that did 1D integrals with an arbitrary mass-dependent kernel (in this case it could be the selection function). Happy to review that PR in CCL if you want me to.

With regards to how to define the selection function: this could be part of a different PR, but would it make sense to factorize the selection function in terms of both the MOR and the observational noise?
E.g. leaving photo-z errors aside, the predicted number of clusters should be roughly of the form:



where  is the selection function you're using (as a function of mass and redshift). It depends on two things in the second line: the MOR () and the noise model (). So, would it make sense at this stage to instead create API for both of these individually and then combine them into a selection function class? As I said, I guess you could first create a toy selection function (as you've done now) and then grow this afterwards.

beckermr:
I assumed that lambda true was not in the model, but it could be. We should ask.

beckermr:
Looking at this more, I think we should leave lambda true out of the model for now. This follows the SRD and is simpler, satisfying our immediate needs.

beckermr:
OK @damonge @elikrause this one is ready for review and merge!

damonge:
Sorry I missed this one (it was in my TODO-list for today). Let me know if you'd like me to look at anything in particular.

beckermr:
Nope we good.

PR #91, opened 2020-03-11
Complete dependencies for example

elisachisari:
Ensures instructions to run example are complete.


beckermr:
Again the requirements are listed in the repo, so we don't need to special-case them in the readmne

PR #92, opened 2020-03-11
Add dependency in example README

elisachisari:
I needed to install `schwimmbad` to run this example. But it wasn't clear to me whether it's needed in other examples, and whether we could replace it by something else. Here are the schwimmbad docs: https://schwimmbad.readthedocs.io/en/latest/


beckermr:
See the requirements here: https://github.com/LSSTDESC/firecrown/blob/master/requirements.txt

My vote is to not merge this since the requirements are listed in the repo.

PR #97, opened 2020-03-27
First draft of changes for cosmosis standalone 0.2.0

joezuntz:
Updates to use newest standalone cosmosis


examples/des_y1_3x2pt/des_y1_3x2pt.yaml
```diff
@@ -1,13 +1,13 @@
 parameters:
   Omega_k: 0.0
   Omega_b: 0.0485
-  Omega_c: 0.2545
+  Omega_c: [0.2, 0.2545, 0.4]
   w0: -1.0
   wa: 0.0
   h: 0.682
-  sigma8: 0.801
+  sigma8: [0.6, 0.801,  0.9]
```
beckermr: Did you decide to punt a single sampled parameters discussion?

joezuntz: Looking at this now.

joezuntz: actually will punt on this for now, as it requires some thought about how things should look and a full rewrite

firecrown/cosmosis/run.py
```diff
@@ -207,16 +208,117 @@ def _make_cosmosis_values(params):
...
+    # Put all the parameters in the data dictionary,
+    # both CCL-related and others, like nuisance params.
+    all_params = data['parameters'].keys()
+    for p in all_params:
+        # string parameters are excluded here, and potentially others
+        if block.has_value('params', p):
+            data['parameters'][p] = block['params', p]
+
+    # Call out to the log likelihood
+    loglike, stats = compute_loglike(cosmo=cosmo, data=data)
```
beckermr: We need to subtract the priors from firecrown here. Also any not used by cosmosis will not get subtracted.

firecrown/cosmosis/run.py
```diff
@@ -207,16 +208,117 @@ def _make_cosmosis_values(params):
...
+    # Now override the varied parameters
+    for p, v in varied_params.items():
+        v = ' '.join(str(x) for x in v)
+        values['params', p] = v
 
-    return cosmosis_values
+    return cosmosis.Inifile(None, override=values)
+
+
+def _make_cosmosis_priors(config):
```
beckermr: Why do we have to redo them for cosmosis. Can't we just use the total output of the priors module like we use for the likelihoods? Also, firecrown will output likelihood+prior. Is there a correction for that I am missing?

requirements.txt
```diff
@@ -10,3 +10,4 @@ emcee>=3,<5
 tqdm
 joblib
 sacc~=0.2
```
beckermr: This should be bumped to 0.3 I think.

firecrown/cosmosis/run.py
```diff
@@ -62,7 +52,7 @@ def run_cosmosis(config, data):
         pool.close()
 
 
-def _make_parallel_pool(cosmosis_config):
+def _make_parallel_pool(config):
```
beckermr: We need to update doc strings for changes in inputs

firecrown/cosmosis/run.py
```diff
@@ -98,7 +89,7 @@ def _make_parallel_pool(cosmosis_config):
     return pool
 
 
-def _make_cosmosis_pipeline(data, values, pool):
+def _make_cosmosis_pipeline(data, ini, values, priors, pool):
```
beckermr: Ditto here for the updated doc string.

README.md
```diff
@@ -36,14 +36,17 @@ See the example in the examples folder for more details.
 
 ## Sampling
 
-To use CosmoSIS to sample cosmological parameters, first install cosmosis-standalone.
-You will need to specify compilers on install, for example:
+To use CosmoSIS to sample cosmological parameters, first install cosmosis-standalone:
 
 ```bash
-CC=gcc-7 CXX=g++-7 FC=gfortran pip install cosmosis-standalone
+pip install cosmosis-standalone
```
beckermr: This should / could mention conda.

firecrown/cosmosis/run.py
```diff
@@ -185,35 +181,163 @@ def _make_cosmosis_params(cosmosis_config):
...
+        # string parameters are excluded here, and potentially others
+        if block.has_value('params', p):
+            data['parameters'][p] = block['params', p]
+
+    # Currently compute_loglike actually computes the posterior
+    # if priors are included. Prevent that from happening since
+    # CosmoSIS is already handling priors
+    if 'priors' in data:
+        data = data.copy()
+        del data['priors']
```
beckermr: This will leave out priors that are not supported by cosmosis but are by firecrown, leaving people pretty confused as to why cosmosis is not applying their prior.

firecrown/cosmosis/run.py
joezuntz: done

firecrown/cosmosis/run.py
joezuntz: done

requirements.txt
joezuntz: done.  shall we add camb to this too?

firecrown/cosmosis/run.py
joezuntz: Any non-supported priors will have triggered an error earlier on in the `_make_cosmosis_priors`

firecrown/cosmosis/run.py
joezuntz: see above.

requirements.txt
beckermr: Nope. Only direct dependencies should be in this file. A dependence of a dependence need not be.

firecrown/cosmosis/run.py
joezuntz: Because there are other things a priors needs to do other than being evaluated, for many samplers - we also need to be able to sample from it and interpolate into a CDF.  Just having them as a single function like now won't do this.

I will look at supporting the generic scipy.stats distributions in cosmosis.

There is a correction below (might be you saw an older version).

firecrown/cosmosis/run.py
beckermr: Ahhhh. Works for me!

beckermr:
The cosmosis tests are being skipped right now. We should turn that off so we run them.

```
firecrown/cosmosis/tests/test_cosmosis.py::test_pool SKIPPED             [ 73%]
firecrown/cosmosis/tests/test_cosmosis.py::test_config SKIPPED           [ 75%]
firecrown/cosmosis/tests/test_cosmosis.py::test_values SKIPPED           [ 78%]
firecrown/cosmosis/tests/test_cosmosis.py::test_pipeline SKIPPED         [ 80%]
firecrown/cosmosis/tests/test_cosmosis.py::test_sampling SKIPPED         [ 82%]
```

joezuntz:
Okay, think this is ready to be a real non-draft PR now.  The tests are enabled whenever cosmosis is installed - I could remove that restriction if you like, now that this puts cosmosis in the install requirements, and now that the whole thing installs much more cleanly.

The latest set of changes enables running Fisher matrices again, and involves minor mods outside the cosmosis module.

I'll merge from master and then convert to a full PR.

joezuntz:
I also took the opportunity to remove some line noise in the YAML file using the `loop_counter` feature in jinja.

beckermr:
Yes, let's remove the skip decorators on the tests to make sure they always run.

examples/cosmicshear/cosmicshear.yaml
```diff
@@ -38,6 +40,17 @@ cosmosis:
...
+  # which parameters to vary in the chains
+  parameters:
+    Omega_c: [0.25, 0.27, 0.32]
+    A_s: [2.0e-9, 2.1e-9, 2.2e-9]
+
+# If no priors are specified then there is
+# an implicit uniform priors on all varied
+# parameters
+priors:
+  module: firecrown.priors
```
beckermr: Why do we need this bit?

firecrown/ccl/two_point.py
```diff
@@ -91,7 +91,7 @@ def compute_loglike(
     else:
         loglike = None
 
-    return loglike, None
+    return loglike, {'data': _data, 'theory': _theory, 'loglike': loglike}
```
beckermr: This change means there needs to be doc change in the function below.

examples/cosmicshear/cosmicshear.yaml
joezuntz: fixed so that this is not required.

firecrown/ccl/two_point.py
joezuntz: done.

.circleci/config.yml
```diff
@@ -63,7 +63,7 @@ jobs:
               "sacc=0.2"
 
             # requires special compilers we don't have
-            # pip install cosmosis-standalone==0.0.9
+            pip install cosmosis-standalone==0.2.2
```
beckermr: You need to do this in conda.

beckermr: So add to the list above

joezuntz: why?  I've never understood when to pip install and when to conda install.

beckermr: pip will pull other deps from pip when they could be in conda, which can end up with messed up envs later. You always use conda first, and then pip, preferably with --no-deps, second. This is a generic recommendation for conda.

joezuntz:
I'd forgotten to update pypi with the newest sacc.  I've done that now - can you re-launch CircleCI?

beckermr:
Looks sacc=0.3 didn't make conda. Fixing now

beckermr:
It needs to be in conda

beckermr:
conda build pushed. We can restart this in about an hour

Issue #106, opened 2020-04-02
force cosmosis outputs to be in the firecrown area


beckermr:
We need to output them to the same dir for consistency.

PR #107, opened 2020-04-03
Make all output files be saved in the same output_ directory

joezuntz:
Addresses #106 

This automatically sets all the cosmosis parameters that involve output paths, so that they will always go in the output directory.  The most obvious output is the chain file, but there are various other auxiliary outputs too that apply to some samplers.

Users are warned if they try to set a parameter which is overridden.

This also involved moving the creation of that output dir to the top level in the firecrown executable.  I also made it use the most up-to-date standard library approach for paths, pathlib.

### DIFF
--- a/bin/firecrown
+++ b/bin/firecrown
@@ -4,7 +4,8 @@ import uuid
 import pprint
 import click
 import jinja2
-
+import pathlib
+import os
 
 import firecrown
 from firecrown.metadata import write_metadata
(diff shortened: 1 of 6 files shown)


beckermr:
Ack. @joezuntz I had fixed part of this in another PR. Would you mind if I ported those changes over to that one?

beckermr:
Also I hesitated to copy the various other files because it might make it hard to restart samplers like multinest. Do you have thoughts on that?

joezuntz:
Do go ahead and port. 

I was assuming we’d need another PR later to address resuming chains, since there’s a couple of different things needed, depending on the sampler.

beckermr:
I'm not familiar with how this works in cosmosis. Do you have docs somewhere?

joezuntz:
I haven't documented it as it's only in the develop/standalone branches, but you just set the parameter `('runtime', 'resume') = 'True'`  and it will do the rest, as long as the output file name is the same.  So should be able to just use the same analysis id and it will work.

joezuntz:
But I'm happy to address later if useful.

beckermr:
Ahhh ok. I'll make another PR after the one I just did. Thanks for the help!

PR #111, opened 2020-04-03
Fix dcos again

beckermr:
I broke them @empEvil. This is the fix.


beckermr:
Also, it turns out I had the firecrown project already. So I added you to that! 🤦‍♂

PR #112, opened 2020-04-14
ENH add ell or theta range for 2pt stats and other binning options

beckermr:
This PR adds bin limit and generation options for 2pt stats. The API is

```yaml
# these cut down any existing bins
ell_or_theta_min: 10
ell_or_theta_max: 100

# these would generate bins to be used
# the min or max options above get applied to these too
# any input SACC data always is used before this block,
# even if it is present
ell_or_theta:
  min: 10
  max: 1000
  n: 10
  binning: 'log'
```

These are extra options to any 2pt stat in the config file.

This PR also renames the old ell_min, etc options to a new block `ell_for_xi` which has keys `min`, `max`, `mid`, `n_log` matching the options in CCL.

### DIFF
--- a/firecrown/ccl/likelihoods/gaussian.py
+++ b/firecrown/ccl/likelihoods/gaussian.py
@@ -46,11 +46,7 @@ class ConstGaussianLogLike(LogLike):
         _sd = sacc_data.copy()
         inds = []
         for stat in self.data_vector:
-            inds.append(
-                _sd.indices(
-                    statistics[stat].sacc_data_type,
-                    statistics[stat].sacc_tracers)
-            )
+            inds.append(statistics[stat].sacc_inds.copy())
         inds = np.concatenate(inds, axis=0)
         cov = np.zeros((len(inds), len(inds)))
         for new_i, old_i in enumerate(inds):
--- a/firecrown/ccl/likelihoods/tdist.py
+++ b/firecrown/ccl/likelihoods/tdist.py
@@ -54,11 +54,7 @@ class TdistLogLike(LogLike):
         _sd = sacc_data.copy()
         inds = []
         for stat in self.data_vector:
-            inds.append(
-                _sd.indices(
-                    statistics[stat].sacc_data_type,
-                    statistics[stat].sacc_tracers)
-            )
+            inds.append(statistics[stat].sacc_inds.copy())
         inds = np.concatenate(inds, axis=0)
         cov = np.zeros((len(inds), len(inds)))
         for new_i, old_i in enumerate(inds):
--- a/firecrown/ccl/statistics/two_point.py
+++ b/firecrown/ccl/statistics/two_point.py
@@ -1,5 +1,7 @@
 import copy
 import functools
+import warnings
+
 import numpy as np
 import pyccl as ccl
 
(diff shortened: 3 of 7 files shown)


firecrown/ccl/likelihoods/gaussian.py
```diff
@@ -46,11 +46,7 @@ def read(self, sacc_data, sources, statistics):
         _sd = sacc_data.copy()
         inds = []
         for stat in self.data_vector:
-            inds.append(
-                _sd.indices(
-                    statistics[stat].sacc_data_type,
-                    statistics[stat].sacc_tracers)
-            )
+            inds.append(statistics[stat].sacc_inds.copy())
```
beckermr, line 49: These now have to be recorded when the statistic is made due to potential cuts.

firecrown/ccl/likelihoods/tdist.py
```diff
@@ -54,11 +54,7 @@ def read(self, sacc_data, sources, statistics):
         _sd = sacc_data.copy()
         inds = []
         for stat in self.data_vector:
-            inds.append(
-                _sd.indices(
-                    statistics[stat].sacc_data_type,
-                    statistics[stat].sacc_tracers)
-            )
+            inds.append(statistics[stat].sacc_inds.copy())
```
beckermr, line 57: ditto here for the tdist

beckermr:
OK @slosar @joezuntz @damonge @chihway 

This PR adds scale cuts (`ell_or_theta_min`, `ell_or_theta_max`) and the ability to generate data at a specified set of input `theta` or `ell` values (`ell_or_theta`).

Ready for a first look!

firecrown/ccl/statistics/two_point.py
```diff
@@ -149,12 +180,51 @@ def read(self, sacc_data, sources):
...
+                "Tracers '%s' have no 2pt data in the SACC file "
+                "and no input ell or theta values were given!" % tracers)
+        elif (
+            self.ell_or_theta is not None
+            and len(_ell_or_theta) > 0
+            and len(_stat) > 0
+        ):
+            warnings.warn(
+                "Tracers '%s' have 2pt data and you have specified `ell_or_theta` "
+                "in the configuration. `ell_or_theta` is being ignored!",
```
damonge, line 197: I think you need `% tracers` here

beckermr:
Yes to all of your questions @damonge. I put the requisite descriptions in the doc strings.

For the last one, we use a grid in ell to compute xi. So the ell_for_xi key is how to specify that grid. Making it bigger or smaller can trade performance for accuracy which is why it is an option.

damonge:
OK, so is the point of `ell_or_theta` that these are the points at which the theory is evaluated, and then it gets interpolated into whatever values are needed to reconstruct the data vector (including window convolution etc.)?

beckermr:
For cell theory computations, we just directly compute it there. For xi, we also do direct computations, but we interpolate in ell using `ell_for_xi` to do the xi computations.

damonge:
Do you mean that there's no window convolution? In that case, shouldn't `ell_or_theta` be determined by whatever binning exists in the sacc file?

beckermr:
We do not support window convolutions in the code currently.

beckermr:
Usually ell or theta is determined by that. However, ppl wanted the ability to do pure theory comps, so I added optional inputs. The SACC file is used by default

damonge:
Oh, I see, got it. OK, gotta go now, but I'll review this over the weekend.

beckermr:
All hell breaks loose. ;)

Would people want a RuntimeError here?

damonge:
`ValueError`? You probably know better than me ;-)

beckermr:
Ahhh sorry. I mean more do you want me to try and correct it by swapping the values? I agree a ValueError is better than RuntimeError.

damonge:
Ah, hehe.
Meh, maybe a check for this specific thing might be good, also maybe not worth the time.
Up to you, you have my LGTM!

beckermr:
OK. I;m gonna merge and move on. Other fish to fry.

PR #123, opened 2020-08-03
Incorporating CMB Lensing Sources and Statistics into Firecrown for 5x2pt 

zhzhuoqi:
I have updated [https://github.com/LSSTDESC/firecrown/blob/5x2/firecrown/ccl/sources/sources.py](url) to include CMB lensing, and updated [https://github.com/LSSTDESC/firecrown/blob/5x2/firecrown/ccl/statistics/two_point.py](url) to calculate two-point statistics of NK (galaxy number count x CMB lensing correlation) and GK (galaxy weak lensing x CMB lensing correlation). So now firecrown should be able to calculate both NK and GK statistics, and I checked these against pyccl, and they give exactly the same output. 

I also tested firecrown MCMC chains (using emcee as sampler) to constrain cosmology on NK and GK. I checked the firecrown results against some MCMC python codes I previously wrote (which I used pyccl and emcee). Using the same priors, the firecrown constraints agree well with my previous results. The NK and GK constraints comparison are shown here. The firecrown constraints are shown in blue, and the previous results of mine are shown in red. 
NK

GK

I also checked other samplers such as maxlike, fisher, and multinest, and they're all able to run through.


firecrown/ccl/sources/sources.py
```diff
@@ -374,3 +374,84 @@ def render(self, cosmo, params, systematics=None):
             systematics[systematic].apply(cosmo, params, self)
 
         self.dndz_interp_ = Akima1DInterpolator(self.z_, self.dndz_)
+
+
+class CMBLSource(Source):
+    """A galaxy cluster source.
```
beckermr, line 380: This doc string is wrong.

firecrown/ccl/sources/sources.py
```diff
@@ -374,3 +374,84 @@ def render(self, cosmo, params, systematics=None):
...
+    systematics : list of str, optional
+        A list of the source-level systematics to apply to the source. The
+        default of `None` implies no systematics.
+
+    Attributes
+    ----------
+    ell_orig : np.ndarray
+        The original ells used in the sacc tracer. Set after the call
+        to `read`.
+    beam_orig : np.ndarray
```
beckermr, line 398: Why are you keeping `ell` and the `beam` here? They are not used iiuic.

firecrown/ccl/sources/sources.py
```diff
@@ -374,3 +374,84 @@ def render(self, cosmo, params, systematics=None):
...
+        Read the data for this source from the SACC file.
+
+        Parameters
+        ----------
+        sacc_data : sacc.Sacc
+            The data in the sacc format.
+        """
+        tracer = sacc_data.get_tracer(self.sacc_tracer)
+        ell = getattr(tracer, 'ell').copy().flatten()
+        beam = getattr(tracer, 'beam').copy().flatten()
```
beckermr, line 429: These `getattr` calls should be written as `ell = tracer.ell.copy().flatten()` etc.

firecrown/ccl/statistics/two_point.py
```diff
@@ -17,6 +17,8 @@
     "galaxy_shear_cl_ee": 'cl',
     "galaxy_shear_xi_minus": 'l-',
     "galaxy_shear_xi_plus": 'l+',
+    "cmbGalaxy_convergenceDensity_xi": 'gg',
+    "cmbGalaxy_convergenceShear_xi_t": 'gl'
```
beckermr, line 21: Please add these options to the docstring for the `TwoPointStatistic` below.

firecrown/ccl/sources/sources.py
```diff
@@ -374,3 +374,84 @@ def render(self, cosmo, params, systematics=None):
...
+        """
+        Read the data for this source from the SACC file.
+
+        Parameters
+        ----------
+        sacc_data : sacc.Sacc
+            The data in the sacc format.
+        """
+        tracer = sacc_data.get_tracer(self.sacc_tracer)
+        ell = getattr(tracer, 'ell').copy().flatten()
```
beckermr, line 428: How does the `ell` here differ from the `ell` pulled out by the two-point class?

firecrown/ccl/sources/sources.py
```diff
@@ -374,3 +374,84 @@ def render(self, cosmo, params, systematics=None):
...
+        cosmo : pyccl.Cosmology
+            A pyccl.Cosmology object.
+        params : dict
+            A dictionary mapping parameter names to their current values.
+        systematics : dict
+            A dictionary mapping systematic names to their objects. The
+            default of `None` corresponds to no systematics.
+        """
+        systematics = systematics or {}
+
```
beckermr, line 451: If you want a systematic to make changes to the beams etc, you'll need to set versions of them via a copy here. Remember the trailing underscore as well.

firecrown/ccl/sources/sources.py
zhzhuoqi: So I hoped to keep a copy of them in case they are changed after calling render. So does it make sense that I keep them here?

beckermr: But what are they used for in the code? What systematic or otherwise would need a copy of `ell` which specifies where things are measured?

firecrown/ccl/sources/sources.py
zhzhuoqi: Yeah, they shouldn't be different. Just to keep a copy

beckermr: This is not the place to keep a copy. That is kept in the two-point object.

beckermr: `ell` should be removed here.

firecrown/ccl/sources/sources.py
zhzhuoqi: Yeah I think it's not needed.

firecrown/ccl/sources/sources.py
zhzhuoqi: Sure

firecrown/ccl/sources/sources.py
zhzhuoqi: But for `beam` does it make sense to read it here and make a copy in `render()`? Or should I directly read and copy it in `render()`?

beckermr: the first. Read it here and then make a copy on render

firecrown/ccl/sources/sources.py
```diff
@@ -425,12 +424,7 @@ def read(self, sacc_data):
             The data in the sacc format.
         """
         tracer = sacc_data.get_tracer(self.sacc_tracer)
-        ell = getattr(tracer, 'ell').copy().flatten()
```
beckermr: Ohhh wait. @damonge would this be the `ell` values for the beam?

beckermr:
Thanks for making the changes. I've asked @damonge to take a look here as I might be wrong about what is in the beams and how we should be treating them.

firecrown/ccl/sources/sources.py
```diff
@@ -374,3 +374,80 @@ def render(self, cosmo, params, systematics=None):
...
+    ----------
+    beam_orig : np.ndarray
+        The original beam used in the sacc tracer before any systematics
+        are applied. Set after a call to `read`.
+    beam_: np.ndarray
+        The beam used for CMB lensing. Set after a call to `render`.
+    scale_ : float
+        The overall scale associated with the source. Set after a call to
+        `render`.
+    tracer_ : `pyccl.WeakLensingTracer`
```
damonge, line 403: Is this right?

firecrown/ccl/sources/sources.py
```diff
@@ -374,3 +374,80 @@ def render(self, cosmo, params, systematics=None):
...
+            A pyccl.Cosmology object.
+        params : dict
+            A dictionary mapping parameter names to their current values.
+        systematics : dict
+            A dictionary mapping systematic names to their objects. The
+            default of `None` corresponds to no systematics.
+        """
+        systematics = systematics or {}
+
+        self.beam_ = self.beam_orig.copy()
```
damonge, line 446: Not sure the beam is being used for anything, actually. In any case, it's unlikely that the CMB lensing maps will be provided with a beam (these usually come from beam-deconvolved CMB maps). They might come with a noise model (e.g. some kind of N_ell), but that's not necessarily relevant here unless you care about the CMB lensing auto-correlation or unless you're generating covariances.

firecrown/ccl/sources/sources.py
```diff
@@ -374,3 +374,80 @@ def render(self, cosmo, params, systematics=None):
...
+        """
+        Read the data for this source from the SACC file.
+
+        Parameters
+        ----------
+        sacc_data : sacc.Sacc
+            The data in the sacc format.
+        """
+        tracer = sacc_data.get_tracer(self.sacc_tracer)
+        beam = tracer.beam.copy().flatten()
```
damonge, line 427: If you're gonna use the beam (and it doesn't seem like you are), just reading the beam doesn't tell you anything, since you don't know what ells it's sampled at.

firecrown/ccl/sources/sources.py
damonge: This seems to be outdated. In any case, indeed beams come with an ell array attached to them. Although as I said below it's unlikely that CMB lensing maps will come with a beam (doesn't hurt to be general though).

firecrown/ccl/sources/sources.py
beckermr: ahhh

beckermr: so I suggested removing the ell for the beam in an earlier comment because I thought it was something else

beckermr: let's remove the beam too then.

firecrown/ccl/sources/sources.py
zhzhuoqi: Ok, Thanks. I'm not planning to do anything like auto-correlation or covariances here. So I'll just remove the beam?

firecrown/ccl/sources/sources.py
zhzhuoqi: Yeah I think so.

firecrown/ccl/sources/sources.py
zhzhuoqi: Thanks, yes the tracer should be a cmb tracer.

firecrown/ccl/sources/sources.py
damonge: Sounds good

firecrown/ccl/sources/sources.py
```diff
@@ -423,9 +423,6 @@ def read(self, sacc_data):
         sacc_data : sacc.Sacc
             The data in the sacc format.
         """
-        tracer = sacc_data.get_tracer(self.sacc_tracer)
```
beckermr: you can't leave this empty - add `pass`

zhzhuoqi: Oh thanks!

zhzhuoqi:
What's a unit test? Sorry I don't know it

beckermr:
See the tests folder for the other sources.

beckermr:
thanks @zhzhuoqi!

PR #125, opened 2021-08-09
Fix priors test

elisachisari:
I was having trouble with one of the tests in Firecrown. It was testing whether number a == number b. I set it so now it tests that the fractional difference is smaller than 1e-10.

### DIFF
--- a/firecrown/tests/test_priors.py
+++ b/firecrown/tests/test_priors.py
@@ -16,9 +16,9 @@ def test_priors():
     loglike, ms, pd, cov, inv_cov, stats = compute_loglike(
         cosmo=None, parameters=parameters, data=data)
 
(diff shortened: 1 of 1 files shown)


marcpaterno:
I think the test might be clearer if we made use of the ‘pytest.approx’ function. This would require importing the pytest module, and then the test could be:

    assert loglike == pytest.approx(
        -0.5 * ((0.3) / 1.0) ** 2
        - 0.5 * np.log(2.0 * np.pi)
        + -0.5 * ((-0.1 - 0.1) / 2.0) ** 2
        - 0.5 * np.log(2.0 * np.pi)
        - np.log(2),
        rel=1e-10)

Is that acceptable to you?

PR #127, opened 2022-03-07
Supernova ayan

am610:
SN Likelihood Integration process developements


vitenti:
PR mixing old and new design. For more up to date PR see " Supernova likelihood #139 ".

PR #129, opened 2022-04-20
Additions for the STAR challenge

tilmantroester:
This is a WIP PR that includes a number of changes necessary for the STAR challenge. I tried to keep the commits reasonably atomic, so they could be merged as separate PRs if that's preferable.  

Things added so far:
- Support for getting the non-linear matter powerspectrum from the data block
- Support for bandpower window functions
- Output the data and theory vectors to the data block for debugging purposes
- Add the pycamb interface to remove the CosmoSIS standard library dependency and thus limit the non-python code to CCL and CAMB only.

### DIFF
--- a/firecrown/connector/cosmosis/likelihood.py
+++ b/firecrown/connector/cosmosis/likelihood.py
@@ -50,7 +50,20 @@ class FirecrownLikelihood:
                 "k": k,
                 "delta_matter:delta_matter": p_k,
             }
-            ccl_args["nonlinear_model"] = "halofit"
+        if sample.has_section("matter_power_nl"):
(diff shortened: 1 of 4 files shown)


joezuntz:
Maybe package the official version of the camb interface, since I've released that now?
https://github.com/joezuntz/cosmosis-standard-library/blob/main/boltzmann/camb/camb_interface.py

vitenti:
The changes here were moved to " Bandpower window function, fisher sampler support #144 " and already merged. There is still the need to have a more lightweight way to have support for CAMB without the full CosmoSIS standard library.

Issue #135, opened 2022-05-16
Refactor for cleaner pylint report


marcpaterno:
This is a step forward toward our goal of having a clean report from pylint.

PR #136, opened 2022-05-16
Refactoring for pylint

marcpaterno:
Closes issue #135. Note that the issue number if the commit message is wrong (sorry).

### DIFF
--- a/firecrown/likelihood/gauss_family/gauss_family.py
+++ b/firecrown/likelihood/gauss_family/gauss_family.py
@@ -15,29 +15,15 @@ from abc import abstractmethod
 
 
 class GaussFamily(Likelihood):
-    """A Gaussian log-likelihood with a constant covariance matrix.
-
-    Parameters
-    ----------
-    statistics : list of Statistic
-        A list of the statistics
-
-    Attributes
-    ----------
-    cov : np.ndarray, shape (n, n)
-        The covariance matrix.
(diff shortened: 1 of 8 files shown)


firecrown/likelihood/gauss_family/gauss_family.py
```diff
@@ -15,28 +15,14 @@
...
-        The (lower triangular) Cholesky decomposition of the covariance matrix.
-    inv_cov : np.ndarray, shape (n, n)
-        The inverse of the covariance matrix.
-
-    Methods
-    -------
-    compute_loglike : compute the log-likelihood
+    """GaussFamily is an abstract class. It is the base class for all likelihoods
+    based on a chi-squared calculation. It provides an implementation of
+    Likelihood.compute_loglike. Derived classes must implement the abstract method
```
vitenti, line 20: Is it not the opposite? GaussFamily implements `compute_chisq` and the derived must implement `compute_loglike` using the value returned by `compute_chisq`?

vitenti (commented):
Just a minor comment on the documentation of GaussFamily.

firecrown/likelihood/gauss_family/gauss_family.py
marcpaterno: You are correct; I'll fix the comment.

PR #144, opened 2022-06-14
Bandpower window function, fisher sampler support

tilmantroester:
This PR adds support for bandpower window functions, getting the non-linear P(k) from CAMB, as well as outputting relevant quantities to the data block to enable the CosmoSIS fisher sampler. This is basically PR #129 but without the contentious inclusion of the pycamb interface in the firecrown repo.

### DIFF
--- a/firecrown/likelihood/gauss_family/statistic/source/source.py
+++ b/firecrown/likelihood/gauss_family/statistic/source/source.py
@@ -81,9 +81,7 @@ class Source(Updatable):
         pass
 
     @final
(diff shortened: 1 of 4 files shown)


marcpaterno:
Sandro and I have gone over this in detail. There are two (new) issues (#147 and #148) reorganizing related code. We are going to clear those issues first, and will then update this pull request to match the new organization. That will mostly involve moving the new blocks of code from this pull request into their new locations.

vitenti (approved):

PR #156, opened 2022-07-01
Added support for relative imports from within likelihood scripts.

vitenti:


vitenti:
@marcpaterno  The current implementation adds the likelihood script to sys.modules using the script name. However, that can lead to name clashing, should we add a prefix? When making a relative import like `from . import x` that will not matter, but if someone uses `import scriptname.x` (which currently works) adding a prefix will require the import to be `import prefix_scriptname.x`.

marcpaterno:
@vitenti I have always been wary of modifying `sys.path` from within code, for the kind of reason you have already raised. However, I think this is a reasonable solution for now. If after some reflection we are able to find a design pattern that does not require modification of `sys.path` we can move to use that pattern at a later date.

marcpaterno (approved):

vitenti:
@marcpaterno Actually, the current solution does not modify sys,path. Actually, I'm following the current wisdom on how to load a script dynamically.

marcpaterno:
I meant to write `sys.modules` rather than `sys.path`. The cautions in the documentation of `sys.modules` has always made me worried about manipulating it. However, as long as we’re careful we should be OK. I may just be being over-cautious.

From: Sandro Dias Pinto Vitenti ***@***.***>
Reply-To: LSSTDESC/firecrown ***@***.***>
Date: Friday, July 1, 2022 at 5:34 PM
To: LSSTDESC/firecrown ***@***.***>
Cc: Marc F Paterno ***@***.***>, Mention ***@***.***>
Subject: Re: [LSSTDESC/firecrown] Added support for relative imports from within likelihood scripts. (PR #156)


@marcpaterno<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_marcpaterno&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=d3p0PLQb2ItFxNVHOOIv2tbqRE3Vbq3EtoL-OopFQXc&m=GuaT5Lrcz6i5zgNbC0rbmX36h9laIhRFSPbYALelp9PfIFlsJ7RsgX127xRf14b5&s=hQ3Abk6aVeteSe7tpE4LcS0mXlYQAJH6tWcnij2_pcw&e=> Actually, the current solution does not modify sys,path. Actually, I'm following the current wisdom on how to load a script dynamically.

PR #157, opened 2022-07-04
First version of Sphinx support. 

vitenti:
It needs several tweaks and all the static documentation.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,0 +1,45 @@
+# Minimal makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line, and also
+# from the environment for the first two.
+SPHINXOPTS    ?= -W
+SPHINXBUILD   ?= sphinx-build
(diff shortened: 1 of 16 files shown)


marcpaterno:
I have verified that I can build the documentation. I had to install 3 packages into my conda environment:
`sphinx`, `sphinx-autodoc-typehints`, and `sphinx_rtd_theme`. With these, I was able to generate HTML documentation. I'm not sure what else is needed to generate the PDF (LaTeX) documentation, or whether we want to support that. 

We should produce a list of required module installations for developers who want to test the generation of documentation, unless we plan on adding all those modules to our "regular" installation. My inclination is to *not* include them in our "regular" installation, just to keep the size down. I am willing to be convinced otherwise.

marcpaterno:
I pushed the wrong button (close with comment, rather than just comment and then merge & close).

marcpaterno (approved):

PR #159, opened 2022-07-06
158 move to GitHub action

vitenti:


vitenti:
Once merged into master we need to remove CircleCI webhooks.

vitenti:
We still need to add caching to the GHA build.

vitenti:
Python's support for final decoration was included in version 3.8. Should we lift the requirement to python>=3.8?

vitenti:
Cobaya example running.

marcpaterno (changes requested):
I have only minor modifications to request:

1. Because new versions of Python deprecate as well as add features, I think we should test on all versions of Python, not just the oldest and most recent we support. This would be 3.8, 3.9 and 3.10 currently and 3.11 as soon as it is released.
2. I would like to alphabetize the dependencies section of environment.yml so that it is easier to find whether or not something is present, or to use "git diff" to find things that are new or removed.

vitenti:
> I have only minor modifications to request:
> 
>     1. Because new versions of Python deprecate as well as add features, I think we should test on all versions of Python, not just the oldest and most recent we support. This would be 3.8, 3.9 and 3.10 currently and 3.11 as soon as it is released.
> 
>     2. I would like to alphabetize the dependencies section of environment.yml so that it is easier to find whether or not something is present, or to use "git diff" to find things that are new or removed.

Done, please take a look.

marcpaterno (approved):

PR #184, opened 2022-10-03
Added state control for updatable objects. 

vitenti:
Added testing and fixed pylint issues (old and new) on tests (including a pylintrc for tests and its inclusion on GHA).

### DIFF
--- a/firecrown/parameters.py
+++ b/firecrown/parameters.py
@@ -3,7 +3,7 @@
 """
 
 from __future__ import annotations
-from typing import List, Dict, Optional
+from typing import Iterable, Dict, Set, Optional
 
 
 def parameter_get_full_name(prefix: Optional[str], param: str) -> str:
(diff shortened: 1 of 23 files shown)


firecrown/parameters.py
```diff
@@ -63,16 +63,16 @@ class RequiredParameters:
     which implements lazy evaluation.
     """
 
-    def __init__(self, params_names: List[str]):
+    def __init__(self, params_names: Iterable[str]):
         """Construct an instance from a list of strings."""
```
marcpaterno, line 67: Comment should either say "from an Iterable yielding strings", or maybe just not describe the type (it is in the type annotations of the signature).

marcpaterno (commented):
There should be two (minor) comments, one on each of two files.

marcpaterno:
This is a comment added *after* pushing the "review" button.

marcpaterno (changes requested):
This is a test of the "request changes" choice for the "Finish your review" control.

tests/likelihood/test_likelihood.py
```diff
@@ -1,13 +1,13 @@
-import pytest
-import os 
+import os
+
 
 from firecrown.likelihood.likelihood import load_likelihood
 
 
 def test_load_likelihood_submodule():
     """The likelihood script shoult be able to load other modules from its
```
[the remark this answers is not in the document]
vitenti: Fixed in the last commits.

firecrown/parameters.py
vitenti: Fixed in the last commit.

PR #188, opened 2022-10-09
Added SN+BAO+CMB examples

am610:
Added `ini` file demonstrating how `SN+BAO+CMB` chains can be produced.


examples/srd_sn/sn_srd_planck_bao.ini
```diff
@@ -0,0 +1,122 @@
...
+[runtime]
+sampler = emcee
+root = ${PWD}
+;; For restarting a chain in a future run
+;; uncomment the following line
+;resume = T
+
+[default]
+fatal_errors = T
+planck_path = ${CSL_DIR}/likelihood/planck2018/baseline/plc_3.0
```
marcpaterno, line 13: `planck_path` is usually used to set paths to the Planck data, as specified elsewhere in the file. I don't see it used -- is that an oversight, or is the `planck_path` parameter not needed?

marcpaterno (changes requested):

marcpaterno (changes requested):
I am not sure I'm interpreting the GitHub web interface correctly, but you may have accidentally committed a file `#generate_sn_data.py#` (perhaps an editor backup file)?

Please check this, and if it is correct, remove the file from the pull request. If it is not correct,, please post a follow-up comment saying this is the case.

am610:
> I am not sure I'm interpreting the GitHub web interface correctly, but you may have accidentally committed a file `#generate_sn_data.py#` (perhaps an editor backup file)?
> 
> Please check this, and if it is correct, remove the file from the pull request. If it is not correct,, please post a follow-up comment saying this is the case.

Please check, I think the `#generate_sn_data.py#` file is successfully removed now.

marcpaterno (approved):

PR #189, opened 2022-10-15
Derived parameters support

vitenti:


marcpaterno (approved):
Documentation for one function updated.

PR #192, opened 2022-10-17
README file formatting correction

am610:
Formatting corrections.

### DIFF
--- a/examples/srd_sn/README.md
+++ b/examples/srd_sn/README.md
@@ -1,61 +1,77 @@
+## Hubble Diagram
+
 The structure of input dataset follows the same pattern as
 Pantheon dataset for cosmosis. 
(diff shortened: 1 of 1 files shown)


marcpaterno (commented):
Please check that description of the LSST DESC formatted file is correct. For example, are the keywords VARNAMES and ROW actually present in the DESC-formatted file?

am610:
> VARNAMES and ROW actually present

Yes this keywords are present. It is so because the same file will be also used in SNANA fast cosmology fitter code : WFIt. Which makes use of this keywords.

am610:
I am not sure, why the fifth check is getting canceled. Can you please advise what should I do?

marcpaterno:
I am approving this PR even though there is a failing check. The failure does not seem to be related to the content of this PR, which changes only a README file.

Issue #193, opened 2022-10-18
IA growth normalization


jablazek:
Is alphag correctly defined given how CCL treats the growth scaling? (use_A_ia = True)

jablazek:
Similarly, in CCL, is the current TATT call double-counting the linear contribution, which assumes a built-in NLA?

tilmantroester:
firecrown: https://github.com/LSSTDESC/firecrown/blob/master/firecrown/likelihood/gauss_family/statistic/source/weak_lensing.py#L182
CCL: https://github.com/LSSTDESC/CCL/blob/master/pyccl/tracers.py#L719
For `alphag==1` the growth term in firecrown drops out but the default in the DES Y1 3x2pt example is -1: https://github.com/LSSTDESC/firecrown/blob/master/examples/des_y1_3x2pt/des_y1_3x2pt_values.ini#L27

tilmantroester:
Looking at the history, `alphag` has used -1 since old firecrown in the examples. @beckermr, do you remember why?

beckermr:
I don't recall exactly. I think I went through the different codes and put in what I thought would be the value that got the right total power of the growth function. I of course could have done that wrong!

tilmantroester:
Right, the issue is that for `use_A_ia==True`, CCL adds the other prefactors (C1, rho_cr, Om), including 1/D, so on the firecrown side it should only set a1 and the redshift scaling, as far as I understand the code.

jpratmarti:
I see, so what you are saying is that firecrown should have `alphag=1` to remove the dependency on D, because it is already implemented in CCL? I think we can do that and rerun everything, however probably this should be changed since it is quite confusing. Ideally if Firecrown should not include any D dependency maybe it should be dropped from the code and we shouldn't need to set alpha_g at all? or else propagate the dependency to CCL. What do you think it is best?

tilmantroester:
My plan is to add functionality to firecrown that allows hiding these parameters from the user (eg #194) but in the short term, setting `alphag=1` is probably the way to go.

jablazek:
CCL has a function to do this bookkeeping for us:
>c_1,c_d,c_2 = pt.translate_IA_norm(cosmo, z, a1=a_1, a1delta=a_d, a2=a_2,
                                   Om_m2_for_c2 = False)

I would suggest we use this function as the baseline, and then give the additional option to add other growth or redshift dependence. But then the default power laws should be zero.

[...]

PR #194, opened 2022-10-18
Allow default parameter values

tilmantroester:
This adds optional default parameters to `ParamsMap.get_from_prefix_param`. This is motivated by issue #193, which could have been avoided by having default values in the code, rather than some ini file.
Before merging, the correct value for `alphag` needs to be confirmed though.

### DIFF
--- a/firecrown/likelihood/gauss_family/statistic/source/weak_lensing.py
+++ b/firecrown/likelihood/gauss_family/statistic/source/weak_lensing.py
@@ -13,6 +13,7 @@ from scipy.interpolate import Akima1DInterpolator
 
 from .source import Source
 from .source import Systematic
+from ..... import parameters
 from .....parameters import (
     ParamsMap,
     RequiredParameters,
@@ -31,7 +32,7 @@ class WeakLensingArgs:
     scale: float
     z: np.ndarray  # pylint: disable-msg=invalid-name
     dndz: np.ndarray
-    ia_bias: Tuple[np.ndarray, np.ndarray]
+    ia_bias: Optional[Tuple[np.ndarray, np.ndarray]]
 
 
 class WeakLensingSystematic(Systematic):
@@ -59,7 +60,7 @@ class MultiplicativeShearBias(WeakLensingSystematic):
     m: float
 
     def __init__(self, sacc_tracer: str):
-        """Create a MultipliciativeShearBias object that uses the named tracer.
+        """Create a MultiplicativeShearBias object that uses the named tracer.
         Parameters
         ----------
         sacc_tracer : The name of the multiplicative bias parameter.
@@ -130,7 +131,7 @@ class LinearAlignmentSystematic(WeakLensingSystematic):
     alphag: float
     z_piv: float
 
-    def __init__(self, sacc_tracer: Optional[str] = None):
+    def __init__(self, sacc_tracer: Optional[str] = None, alphag=1.0):
         """Create a LinearAlignmentSystematic object, using the specified
         tracer name.
 
@@ -147,28 +148,22 @@ class LinearAlignmentSystematic(WeakLensingSystematic):
         """
         super().__init__()
 
+        self.ia_bias = parameters.create()
+        self.alphaz = parameters.create()
+        self.alphag = parameters.create(alphag)
+        self.z_piv = parameters.create()
+
         self.sacc_tracer = sacc_tracer
 
     @final
     def _update(self, params: ParamsMap):
-        self.ia_bias = params.get_from_prefix_param(self.sacc_tracer, "ia_bias")
-        self.alphaz = params.get_from_prefix_param(self.sacc_tracer, "alphaz")
-        self.alphag = params.get_from_prefix_param(
-            self.sacc_tracer, "alphag", default=1.0
-        )
-        self.z_piv = params.get_from_prefix_param(self.sacc_tracer, "z_piv")
+        pass
 
     @final
     def _reset(self) -> None:
         """Reset this systematic.
 
         This implementation has nothing to do."""
-
-    @final
-    def required_parameters(self) -> RequiredParameters:
-        return RequiredParameters(
-            [parameter_get_full_name(self.sacc_tracer, pn) for pn in self.params_names]
-        )
 
     @final
     def _get_derived_parameters(self) -> DerivedParameterCollection:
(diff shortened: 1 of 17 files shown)


marcpaterno:
Cobaya requires that the names of all parameters be specified. We rely on the `RequiredParameters`  object containing a list of names to let us know what to tell Cobaya is required. The introduction of default into this system would break this; we would not know whether to tell Cobaya that `alphag` is required or not.

In addition, it looks to us like this design is vulnerable to an easy-to-make user error. If a user misspells a parameter name (e.g. writes `alpha_g` rather than `alphag`, the extraneous name `alpha_g` will be silently ignored, and the default value for `alphag` will be used.

Please take a look at how the optional `mag_bias` is handed in `NumberCounts`. We are using name-only arguments in constructors, sometimes with default values, to help make sure that an accidental misspelling of a parameter name in a configuration will not pass quietly, and lead to unwanted behavior. Can a similar design meet your needs here?

tilmantroester:
> In addition, it looks to us like this design is vulnerable to an easy-to-make user error. If a user misspells a parameter name (e.g. writes alpha_g rather than alphag, the extraneous name alpha_g will be silently ignored, and the default value for alphag will be used.

For cosmosis there is a flag that raises an error when a parameter is never accessed IIRC.

I'd argue that having to include parameters in a configurations file that the user doesn't need or understand but are necessary for a general form of a model is more prone to user error. For an explicit example, see issue #193.

If cobaya doesn't support the concept of optional parameters then the implementation might need to change how the required parameters are set up. But we need some way to implement systematics where a subset of the parameters have default values that can however be varied for specific analyses.

Here's an implementation that works with cobaya but is a bit verbose for my taste. Setting `ia_a_d=None` makes it a required parameter again. But I haven't checked what happens in the case of per-source parameters. It would be nice if populating the `self.params_with_defaults` list were automatic.
```python
def __init__(self, sacc_tracer: Optional[str] = None, ia_a_d=0.7):
        super().__init__()
        self.sacc_tracer = sacc_tracer
        self.params_with_defaults = []
        if ia_a_d is not None:
            self.params_with_defaults += ["ia_a_d"]
            self.a_d = ia_a_d

    @final
    def _update(self, params: ParamsMap):
        self.a_1 = params.get_from_prefix_param(self.sacc_tracer, "ia_a_1")
        self.a_2 = params.get_from_prefix_param(self.sacc_tracer, "ia_a_2")
        if "ia_a_d" not in self.params_with_defaults:
            self.a_d = params.get_from_prefix_param(self.sacc_tracer, "ia_a_d")

    @final
    def required_parameters(self) -> RequiredParameters:
        return RequiredParameters(
            [parameter_get_full_name(self.sacc_tracer, pn) for pn in self.params_names if pn not in self.params_with_defaults]
        )
```

marcpaterno:
We have tried to make a general solution. We have provided an example below. In this design:

1. For clarity: a "required parameter" means that parameter is required to be read from the sample (e.g. the CosmoSIS `datablock`).
2.  Some parameters are always required; in the code below `ia_a_1` and `ia_a_2` are of this nature.
3. Other parameters are required or not, under the user's control (the user is the author of the likelihood factory function which calls the constructor of the Updatable in question). It is the value passed to the constructor that is used to make the decision. The value may be there for either of two reasons: (1) because the user specified it or (2) because the user did not specify it but the author of the Updatable specified a default.
4. The author of the Updatable class can provide default values, either `None` or some `float` value, or not provide a default, as seems most fitting.
5. The value actually used by the constructor at runtime determines whether the parameter in question is required or not:
    1. if the value is `None` the parameter is required; it must be provided by the sampler, and a runtime error will happen if it is not.
    2. if the value is any `float` value the parameter is not required; it must not be provided by the sample, and a runtime error will happen if it is provided. 

Here is our proposed replacement for your code above:

```python
def __init__(self, sacc_tracer: Optional[str] = None, *, ia_a_d=0.7, ia_x=None, ia_y):
    super().__init__()
    self.sacc_tracer = sacc_tracer
    # We hope to make the next line more automated...
    self.required_parameter_names = ["ia_a_1", "ia_a_2"] # always required

    # We hope to be able to automate the following through introspection    
    if ia_a_d is None:
        self.required_parameter_names += ["ia_a_d"]
    if ia_x is None:
        self.required_parameter_names += ["ia_x"]
    if ia_y is None:
        self.required_parameter_names += ["ia_y"]

@final
def _update(self, params: ParamsMap):
    # We may be able to push this implementation into the framework
    for name in required_parameter_names:
        setattr(self, name, params.get_from_prefix_param(self.sacc_tracer, name))
	
@final
def required_parameters(self) -> RequiredParameters:
   # We may be able to push this implementation into the framework
    return RequiredParameters([parameter_get_full_name(self.sacc_tracer, pn) 
                               for pn
			       in self.required_parameter_names])
 )
```

Does this meet your needs?

tilmantroester:
Looks good to me!

vitenti:
@marcpaterno @tilmantroester , this PR ready to review. @am610 Ayan, please note that I had to include the sacc tracer into the supernova parameter M, the default behavior to a Updatable parameter name is to include the sacc tracer as its prefix to avoid name clashing between different observables/likelihoods. Now this must be called sn_ddf_sample_m (or sn_ddf_sample_M) instead of simply M.

firecrown/likelihood/gauss_family/statistic/source/weak_lensing.py
```diff
@@ -49,43 +49,33 @@ class MultiplicativeShearBias(WeakLensingSystematic):
...
         ----------
         sacc_tracer : The name of the multiplicative bias parameter.
         """
         super().__init__()
 
+        self.mult_bias = parameters.create()
         self.sacc_tracer = sacc_tracer
 
     @final
     def _update(self, params: ParamsMap):
```
tilmantroester, line 66: The `_update`, `_reset`, `_required_parameters` methods (and `_get_derived_parameters` to a lesser degree) add a lot of boiler plate but in most cases don't actually do anything any more, as shown by the many `"This implementation has nothing to do."`
I'd like to suggest removing the `@abstractmethod` for these in `Updatable`, or at least move the default implementations to places like `Systematic` and `Source`. I imagine the most common contributions to firecrown will be in the form of new systematics by users that neither understand nor care about the very neat internals you've come up with. I believe that allowing users to write a simple `Systematics` class with just an `__init__` and `apply` method makes for a more welcoming environment and readable code than having to copy in a bunch of empty `_*` methods that most users won't ever need to change. 

The existence of these methods should be documented for advanced users but even knowing what these methods do, I personally struggle to come up with a use case right now for anything but `_get_derived_parameters`.

marcpaterno (approved):

PR #196, opened 2022-10-19
Fix min->minimum in two_point.py

tilmantroester:
A recent refactor broke two_point.py when using bandpowers.


marcpaterno (approved):
Thank you for finding and fixing this error.

Issue #174, opened 2022-08-05
Treatment of Systematics Effects in 3x2 point statistics


chrgeorgiou:
Currently, the only way to treat systematic effects in 3x2 point statistics is through modifying the variables of a ```ccl.Tracer``` object. For systematic effects that require the calculation of a specific power spectrum that is different from the standard matter power spectrum (e.g. extended intrinsic alignment models such as TATT and Halo model, treatment of baryons, non-linear galaxy bias but also maybe extended cosmologies) the current architecture does not allow their treatment. Below is a code that demonstrates what this would look like with the TATT intrinsic alignment model using CCL. 

```
import numpy as np
import pyccl as ccl
import pyccl.nl_pt as pt
cosmo = ccl.Cosmology(Omega_c=0.27, Omega_b=0.045, h=0.67, A_s=2.1e-9, n_s=0.96)
ell = np.unique(np.geomspace(2,1000,100).astype(int)).astype(float)
# redshift distribution
z = np.linspace(0, 1.5, 1024)
nz = np.exp(-((z-0.7)/0.1)**2)

# Systematics effect is controlled with parameters c_1, c_2, c_d.
# Biases for IAs. Will be converted to the input c_IA values below.
a_1 = 1.
a_2 = 0.5
a_d = 0.5
c_1, c_d, c_2 = pt.translate_IA_norm(cosmo, z, a1=a_1, a1delta=a_d, a2=a_2, Om_m2_for_c2 = False)

# Code that creates a Pk2D object:
ptc = pt.PTCalculator(with_NC=True, with_IA=True,
                      log10k_min=-4, log10k_max=2, nk_per_decade=20)
ptt_i = pt.PTIntrinsicAlignmentTracer(c1=(z,c_1), c2=(z,c_2), cdelta=(z,c_d))
ptt_m = pt.PTMatterTracer()
# IAs x matter
pk_im = pt.get_pt_pk2d(cosmo, ptt_i, tracer2=ptt_m, ptc=ptc)

# Code that computes effect from IA using that Pk2D object
t_lens = ccl.WeakLensingTracer(cosmo, dndz=(z, nz))
t_ia = ccl.WeakLensingTracer(cosmo, dndz=(z, nz), has_shear=False, ia_bias=(z, np.ones_like(z)), use_A_ia=False)
cl_GI = ccl.angular_cl(cosmo, t_lens, t_ia, ell, p_of_k_a=pk_im)
# The weak gravitational lensing power spectrum
[...]

PR #198, opened 2022-10-21
Add support for non-linear models

tilmantroester:
This is a draft implementation to address #174 based on discussions with @jablazek, @paulrogozenski, @chrgeorgiou. It currently supports perturbative IA and galaxy bias models but it should be extendable to halo-model-based models. An example (`examples/des_y1_3x2pt/des_y1_3x2pt_PT.py`) showcases this functionality and compares the results to a direct CCL calculation of the same quantities. They seem to match well!

Allowing support for non-linear quantities requires extensive changes to how firecrown handles two-point statistics and passes around cosmology objects. The main changes are

- Add a `CosmologyBundle` class that bundles the `ccl.Cosmology` object with associated data, such as PT calculators. This class gets passed around as `cosmo` instead of the `ccl.Cosmology` object downstream of the top-level `Likelihood` call.
- Add a `TracerBundle` class that bundles a `ccl.Tracer` object with information about the associated 3D field and `ccl.nl_pt.PTTracers`, where relevant.
- Add a `CosmoSystematicsLikelihood` likelihood class that can account for cosmology-level systematics. The concrete example here is the new class `PTSystematic`, which populates the PT workspaces and computes the basic perturbative quantities that are later used in `TwoPoint.compute`
- Changed how `Source.create_tracer` and `TwoPoint.compute` work to deal with the multiple tracers that are now required per `Source`
- Add a `TattAlignmentSystematic` and `PTNonLinearBiasSystematic` systematic that make use of all this new functionality
- Some refactoring in `number_counts.py` to more generally deal with magnification bias

A plot showing the agreement with CCL:
pt_cls
This PR currently uses the `pt_enh` branch of CCL, since it required some extra functionality there.


tilmantroester:
@marcpaterno, @vitenti Do you have comments or suggestions on the overall structure of the implementation? There's still quite some clean-up to do but I want to make sure you're on board with the design before investing more time in writing and changing the code.

vitenti:
@tilmantroester yesterday Marc and I finished the design and I did most of the implementation this morning, I'll push it so we can discuss if it fits all the needs.

vitenti:
@tilmantroester We are still reviewing the code structure and figuring out how to better integrate with design we were developing for the toolset.

marcpaterno:
@tilmantroester We have just cut firecrown release v1.3.0, containing (among other things) the new version of the likelihood "factory script".

Would you like this code to make it into a v1.4.0, which we could cut in the next week or two? As of our last meeting, our plan was to release this code to be used in the forecasting project (DESC project 291) and to make the generalization/refactoring we discussed as a task to be done after that release.

tilmantroester:
I'm busy until next week but after that I'll try to get this ready to be reviewed.

tilmantroester:
This should now be ready to be looked at again. I've added a unit test which automatises the example scripts. This should now also handle combinations of PT and non-PT tracers, following a similar approach to @paulrogozenski.

One thing to discuss is whether to use the `firecrown.likelihood.Cosmology` everywhere. Right now the methods of `Likelihood` still take `pyccl.Cosmology` arguments, and only for certain downstream cases `Statistic.compute`, the firecrown `Cosmology` class is used.

tilmantroester:
This is now finally passing all tests.

vitenti (approved):

Issue #175, opened 2022-08-12
Not exactly an installation issue, but related to confusing error messages when running Firecrown for the first time without having set {FIRECROWN_DIR} to point to the correct directory. The final error message seen by the user is distracting from the actual issue.


marcpaterno:
Not exactly an installation issue, but related to confusing error messages when running Firecrown for the first time without having set {FIRECROWN_DIR} to point to the correct directory. The final error message seen by the user is distracting from the actual issue.

Here when running "cobaya-run cobaya_evaluate.yaml" (but apparently also applicable to cosmosis):

<...>
```
FileNotFoundError: [Errno 2] No such file or directory: '${FIRECROWN_DIR}/examples/des_y1_3x2pt/des_y1_3x2pt_sacc_data.fits'
The above exception was the direct cause of the following exception:
```
<...>
```
 File "/Users/agne/opt/anaconda3/envs/fc/lib/python3.10/site-packages/firecrown/likelihood/likelihood.py", line 105, in load_likelihood
    raise ImportError(f"{e.strerror}: {filename}") from e
ImportError: No such file or directory: des_y1_3x2pt.py
```

_Originally posted by @agnesemenaite in https://github.com/LSSTDESC/firecrown/issues/172#issuecomment-1206740158_

marcpaterno:
This is a problem caused by the fact that we have code that can fail being executed at `import` time. We should change the design of our "factory function" scripts to be instead python modules that declare a factory function (that returns the likelihood object). This will make it easier to detect this kind of failure and to product a meaningful error message.

tilmantroester:
I think the issue is that `FileNotFoundError` gets caught and then re-raised as an unhelpful `ImportError` here: https://github.com/LSSTDESC/firecrown/blob/master/firecrown/likelihood/likelihood.py#L104
See the MWE in the other issue I opened on this: #173 
Other exceptions being raised just show up as one would expect, it's just `FileNotFoundError` (which are very common) that get obfuscated.

vitenti:
Fixed by PR #207

Issue #173, opened 2022-08-05
Unhelpful error when likelihood file raises exception


tilmantroester:
If the .py file raises an exception, the last exception being shown is an `ImportError` with the name of the likelihood .py file, which obscures the original exception (`FileNotFoundError`, in this case).

`mwe_likelihood.py`:
```python
from firecrown.likelihood.gauss_family.gaussian import ConstGaussian

raise FileNotFoundError

likelihood = ConstGaussian()
```

```
[output] Output to be read-from/written-into folder '.', with prefix 'cobaya_evaluate_output'
[output] Found existing info files with the requested output prefix: 'cobaya_evaluate_output'
[camb] `camb` module loaded successfully from /Users/yooken/Codes/miniforge3/envs/firecrown/lib/python3.10/site-packages/camb
[exception handler] ---------------------------------------

Traceback (most recent call last):
  File "/Users/yooken/Research/LSST/dev/firecrown/firecrown/likelihood/likelihood.py", line 96, in load_likelihood
    spec.loader.exec_module(mod)
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/Users/yooken/Research/LSST/dev/firecrown/broken_examples/fileerror/mwe_likelihood.py", line 3, in <module>
    raise FileNotFoundError
FileNotFoundError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/yooken/Codes/miniforge3/envs/firecrown/bin/cobaya-run", line 8, in <module>
    sys.exit(run_script())
  File "/Users/yooken/Codes/miniforge3/envs/firecrown/lib/python3.10/site-packages/cobaya/run.py", line 225, in run_script
    run(info, **arguments.__dict__)
  File "/Users/yooken/Codes/miniforge3/envs/firecrown/lib/python3.10/site-packages/cobaya/run.py", line 145, in run
    with Model(updated_info["params"], updated_info["likelihood"],
  File "/Users/yooken/Codes/miniforge3/envs/firecrown/lib/python3.10/site-packages/cobaya/model.py", line 237, in __init__
    self.likelihood = LikelihoodCollection(info_likelihood, theory=self.theory,
  File "/Users/yooken/Codes/miniforge3/envs/firecrown/lib/python3.10/site-packages/cobaya/likelihood.py", line 262, in __init__
[...]

PR #201, opened 2022-10-25
Pass FileNotFoundError exceptions in likelihood through

tilmantroester:
If a likelihood raises a `FileNotFoundError`, for example due to a missing `sacc` file, that exception gets reraised as an `ImportError` with the likelihood file name and buries the actual missing file way up in the trace. See #173 and #175.

This PR removes the `try...except` block around `spec.load.exec_module`. `FileNotFoundError` now gets passed through correctly. If the actual module cannot be found, that still raises an informative exception.


marcpaterno:
We plan to update the current likelihood factory scripts to provide likelihood factory functions. Until we do that, this is already a step in the right direction.

Issue #244, opened 2023-03-21
A_s vs sigma_8


elisachisari:
In the cosmicshear example, within cosmicshear_values.ini, it seems both sigma_8 and A_s are being fixed. How does it work in the current implementation? Only one should be necessary, so why are both specified?

joezuntz:
You also need to rename the parameter to `sigma_8_input` (added underscore).

PR #250, opened 2023-03-28
A s vs sigma 8

marcpaterno:
This closes issue #244 .


vitenti (approved):

vitenti:
@marcpaterno I think we should update all other examples as I explained in the discussion on slack. I can do it sometime this week but fell free to do it if you have the time.

tilmantroester:
We might want to add a note somewhere about the caveats using sigma8 as an input, such as the extra compute time (because CAMB is basically run twice) and that the output sigma8 can differ from the input slightly (because of aggressive accuracy settings on the first run of CAMB).

Issue #251, opened 2023-03-30
A_s vs sigma_8 in all CosmoSIS examples


marcpaterno:
As was noted in issue #244 and in the PR #250 that closed it, all other CosmoSIS examples that set both `A_s` and `sigma_8` should be updated.

PR #253, opened 2023-03-30
Fix all the CosmoSIS examples, remove A_s

marcpaterno:
This closes issue #251


vitenti (approved):

tilmantroester:
I'd advocate against changing to only using sigma_8 everywhere. There are good reasons to not sample sigma_8 and removing all references to As from the examples might hide this option from the user.
Instead, the sigma8_input should be removed in at least some of the examples, or a comment added to alert user that both As and sigma8 can be sampled, with the caveats that come along with using sigma8.

vitenti:
@tilmantroester I agree, I think that's probably better to sample A_s directly. Moreover, we were obtaining weird results when using sigma8 directly. I'll investigate this first, but then, we can roll back these modifications.

PR #273, opened 2023-05-16
Create figure in current directory

marcpaterno:
This fixes the issue by creating the figure in the current directory, which was simpler than creating a subdirectory.

### DIFF
--- a/examples/des_y1_3x2pt/des_y1_3x2pt_PT.py
+++ b/examples/des_y1_3x2pt/des_y1_3x2pt_PT.py
@@ -295,8 +295,7 @@ def run_likelihood() -> None:
     [a.legend(fontsize="small") for a in ax]
 
     fig.suptitle("PT Cls, including IA, galaxy bias, magnification")
-    with open("pt_cls.png", "w+b") as png_file:
-        fig.savefig(png_file, facecolor="white", dpi=300)
+    fig.savefig("pt_cls.png", facecolor="white", dpi=300)
 
     plt.show()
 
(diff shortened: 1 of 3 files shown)


examples/des_y1_3x2pt/des_y1_3x2pt_PT.py
```diff
@@ -295,7 +295,8 @@ def run_likelihood() -> None:
     [a.legend(fontsize="small") for a in ax]
 
     fig.suptitle("PT Cls, including IA, galaxy bias, magnification")
-    fig.savefig("plots/pt_cls.png", facecolor="white", dpi=300)
+    with open("pt_cls.png", "w+b") as png_file:
+        fig.savefig(png_file, facecolor="white", dpi=300)
```
tilmantroester, line 299: Why the context manager instead of using the file name directly in `savefig`?

marcpaterno: Habit. I prefer to make it obvious the resource is “cleaned up” (in this case, the file closed) rather than having a reader need to know whether the routine being called does the right thing.

tilmantroester: If it's just here that fine but generally I'd argue having a context manager for `savefig` is overkill and likely to confuse users (what happens if you forget the `"b"`?). Might as well throw in a `os.makedirs("plots", exists_ok=True)` to deal with the subdir in that case.

tilmantroester (approved):

PR #287, opened 2023-06-19
Created ClusterMassTrue and BinLogMTracer for true mass analysis

eduardojsbarroso:
Created the ClusterMassTrue object and its needed tracer to perform the cluster counts analysis when we have the true masses of te clusters (i.e. from simulations).


eduardojsbarroso:
Pylint is evaluatting the code at 9.99 because I added two local variables in `tests/test_cluster.py` in the fixture function `fixture_cluster_objects`. Should I split this function in two to change this? Or we can also set the maximum of local variables to 17.

Issue #305, opened 2023-08-08
Move to SACC 0.11


marcpaterno:
Because `sacc` v 0.11 contains the tracer support code in `firecrown/sacc_support.py`, we should remove `sacc_support.py`, and adjust any code to use SACC’s version of those classes.

Issue #295, opened 2023-07-18
Remove Sacc_support


eduardojsbarroso:
Since the new tracers have been merged in SACC, we need to remove from firecrown and update all the notebooks

vitenti:
Done in #306

PR #306, opened 2023-08-08
Move to sacc 011

marcpaterno:
This will close issue #305


mattkwiecien:
This will also close #295 👍

vitenti (approved):
Coverage was reduced because code was deleted but no new uncovered code was added.

PR #316, opened 2023-09-20
Create a parameter prefix for updatables

vitenti:


firecrown/updatable.py
```diff
@@ -57,14 +57,22 @@ class Updatable(ABC):
 
     """
 
-    def __init__(self) -> None:
-        """Updatable initialization."""
+    def __init__(self, parameter_prefix: Optional[str] = None) -> None:
+        """Updatable initialization.
+
+        :param prefix: prefix for all parameters in this Updatable
```
marcpaterno, line 63: We should fix the documentation so that it matches the argument name “parameter_prefix” (or change the code to match the comment).

marcpaterno (approved):

PR #319, opened 2023-10-04
Fix crash on S8 sampling in example

elisachisari:
Sampling over S8 in cosmic shear example file was broken. This is a quick fix.

### DIFF
--- a/examples/cosmicshear/cosmicshear.ini
+++ b/examples/cosmicshear/cosmicshear.ini
@@ -24,14 +24,17 @@ file = ${CSL_DIR}/utility/consistency/consistency_interface.py
 [camb]
 file = ${CSL_DIR}/boltzmann/camb/camb_interface.py
 mode = all
-lmax = 2500
+lmax = 2000
 feedback = 0
 zmin = 0.0
-zmax = 4.00
+zmax = 3.00
 nz = 100
 kmin = 1e-4
 kmax = 50.0
 nk = 1000
+
+[sigma8_rescale]
+file = ${CSL_DIR}/utility/sample_sigma8/sigma8_rescale.py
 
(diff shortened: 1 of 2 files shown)


marcpaterno:
Elisa, can you tell us what version of the CosmoSIS Standard Library you are using? This seems to be reverting to an older version of the CSL; we would prefer, if possible, to move everyone to using the newer version.

If we need to support multiple inconsistent versions of CosmoSIS and CSL (and perhaps also NumCosmo, and Cobaya), we'll need to discuss how we should do so. We had the discussion for CCL, where the conclusion was ... don't support the older versions.

marcpaterno:
@joezuntz In the current version of CosmoSIS, is the `sigma8_rescale` module deprecated? If not, if someone both sets `sigma8` and uses the `sigma8_rescale` module, does anything weird happen? It appears that `sigma8_rescale` reads `sigma8_input`, which is already marked as deprecated.

tilmantroester:
As of v2.5 of the CSL (and v3.3.2 of cobaya), sigma8 can be used as an input to CAMB without approximations or notable performance hits, so earlier methods like `sigma8_rescale` or `sigma8toAs` aren't required any more.

marcpaterno:
These changes would revert the example to use a configuration that was required by older versions of CosmoSIS. While it also works with the newer version of CosmoSIS the result is:
1. less efficient, and 
2. does not deal correctly with non-linear power spectra coming from CAMB.

We would prefer to reject this PR (to not revert the configuration) unless there is a need to retain examples supporting the older version of CosmoSIS.

Is there such a need?

elisachisari:
Tried updating the environment to work with latest CSL and the sampling over S8 works out of the box. So I'm happy to close this.

PR #329, opened 2023-10-17
Power spectrum modifier systematic

tilmantroester:
This PR adds functionality to enable more-or-less arbitrary modifications to the 3D power spectra. The primary motivation is the inclusion of baryons but the API is general. For example, this could be used to compute the non-linear power spectrum in general.

1. The `PkSystematic(field="delta_matter_baryons")` specifies which field (3D power spectrum) the tracer should use when computing the Cls. Usage shown here: https://github.com/LSSTDESC/firecrown/blob/pk_modifier/examples/des_y1_3x2pt/des_y1_cosmic_shear_pk_modifier.py#L98
2. `PowerspectrumModifier` is the API where the heavy lifting happens. This implements an attribute `name` and a `compute_p_of_k_z` method. A list of `PowerspectrumModifier` gets passed to `ModelingTools` during instantiation. During `ModelingTools.prepare`, `compute_p_of_k_z` of the `PowerspectrumModifier`s gets called and the resulting `Pk2D` objects get added to the power spectrum library  as `name`. A minimal example implementation of the van Daalen+2019 model in CCL is here: https://github.com/LSSTDESC/firecrown/blob/pk_modifier/examples/des_y1_3x2pt/des_y1_cosmic_shear_pk_modifier.py#L26, with instantiation here: https://github.com/LSSTDESC/firecrown/blob/pk_modifier/examples/des_y1_3x2pt/des_y1_cosmic_shear_pk_modifier.py#L55

vD19_baryons

This is a draft implementation so far and comments on the API are very welcome.

One implementation detail is whether to reset the computed power spectra in `ModelingTools`. Are there use cases where the power spectra are cosmology-independent?

Todo:
- [ ] Check `update`, `required_parameters`, etc are called appropriately in the connectors
- [ ] Add tests
- [ ] Add doc string


tilmantroester:
@marcpaterno @vitenti Do you have any comments on this implementation?

vitenti:
Hi @tilmantroester, we discussed this during our working session today. Our main question is why use a new member of the `tracer_args` dataclass instead of just adding a new property `field` to the `WeakLensing` and `NumberCounts` sources, or even to the base class `SourceGalaxy`?

The current implementation leaves us wondering whether a second systematic might alter the `.field` member, potentially leading to different behavior. While it's challenging to create foolproof code, in this situation, it could result in challenging-to-detect bugs.

tilmantroester:
I see your point, setting the default `field` as a `SourceGalaxy` attribute would also work. My reasoning was that having it as a systematic that modifies `tracer_args` is that it follows the same structure as the other systematics, i.e., there's only one place to add "systematics", and users don't need to remember which effects are "systematics" and which are set in the `WeakLensing` constructor. Note that other systematics, for example the PT ones, implicitly set the `field` too.

I think that `tracer_args` being potentially overwritten is a general issue with the current implementation of systematics. Having the option to issue warnings if any of the `tracer_args` attributes get modified more than once would be useful once we move more towards production usage of more complicated firecrown likelihoods.

vitenti:
Hi @tilmantroester , everything seems fine, we reviewed it and agree that's in good shape. Please, tell us if you agree so we can merge it. I added tests and docs, please take a look to see if you agree and think everything is done.

tilmantroester:
Looks good to me. Thanks for added the tests and docs! I just updated the docs in the example. Should #336 be merged first?

vitenti:
@tilmantroester  Since #336 decreases the coverage, we can just merge this one and then close #336. Or we can add tests for #336, I didn't have a good example of window function usage to create a test. If you have a simple test to cover the new added lines, please add that to #336 then we can merge it first and then this one.

tilmantroester:
Ok, let's merge this one then. Tests for the window functions would be good to have in general but I don't have a small sacc file on hand for testing purposes.

vitenti (approved):

PR #334, opened 2023-10-20
Renamed create to register_new_updatable_parameter.

vitenti:


vitenti:
@tilmantroester tell me if that's what you had in mind. I kept the old `create` function as a deprecated function to be removed in the next major release.

marcpaterno (approved):
I think the new name is a clear improvement.

tilmantroester:
LGTM!

PR #336, opened 2023-10-20
Fix window function normalisation

tilmantroester:
This is a quick fix to ensure the bandpower window functions are normalised to 1.


vitenti:
Fixed in #329

PR #337, opened 2023-10-20
Update installation for support m2 mac

marcpaterno:
This PR has many changes to the documentation and to the recommended development  workflow. It includes now support for Apple Silicon (M1 and M2) Macs.


docs/dev-notes.rst
```diff
@@ -5,15 +5,17 @@ Developer Notes
...
-To run the current version, one needs to *build* (but not yet install) Firecrown.
-In this directory, run:
+To run the development version, one needs to do an *editable installation* of firecrown.
+This will create an entry in the conda environment that knows about the local code.
+You must have the `:bash:firecrown_developer` conda environment activated before running this installation.
 
-    python3 setup.py build
+Note that we use the :bash:`--no-deps` flag to prevent the installation from accidentally taking in any new dependencies through  :bash:`pip`.
+If the installation fails because of the lack of a dependency, install that dependency using  :bash:`conda` and not  :bash:`pip`.
+If you find a dpendency on a package not available through :bash:`conda` please file an issue on the issue tracker.
```
vitenti, line 14: typo on *dependency*

docs/dev-notes.rst
```diff
@@ -5,15 +5,17 @@ Developer Notes
...
 
-This will put modules into the subdirectory :bash:`build/lib`.
-Set the environment variable :bash:`FIRECROWN_DIR` to the full path to that directory.
-Set the environment variable :bash:`FIRECROWN_EXAMPLES_DIR` to be the full path to the :bash:`examples` subdirectory.
-
-These environment variables are needed by the example :bash:`ini` files.
-The directory :bash:`$FIRECROWN_DIR` should also be on :bash:`PYTHONPATH`.
+.. code:: bash
+    
+    python -m pip install --no-deps --editable ${PWD}
```
vitenti, line 18: Should it be strict? https://setuptools.pypa.io/en/latest/userguide/development_mode.html#strict-editable-installs

vitenti (commented):
The primary question here is whether to recommend editable or strictly editable installations. In the second scenario, strict editable installations necessitate reinstallation whenever changes are applied but offer a closer simulation to a real installation.

docs/dev-notes.rst
tilmantroester: I feel like strict-editable would lead to more confusion. I imagine the people that would use an editable install want quick iterations while developing and know to test it as a normal install at the end.

vitenti: Yes, we agree to keep just "editable"

vitenti (approved):

Issue #340, opened 2023-10-26
Support loading likelihoods from python packages


tilmantroester:
It would be useful to support loading the `build_likelihood` factory from python packages, rather than file names.
For example, to support usage like `firecrownIni: desc_baryon_challenge.likelihoods.firecrown_likelihood`.
Or maybe better more explicitly: `firecrown_likelihood_factory: desc_baryon_challenge.likelihoods.firecrown_likelihood.build_likelihood`

With the option to pass `build_parameters`, this should allow quite generic likelihood factories that can be shipped centrally in a python package.

A specific use case is the baryon challenge: https://github.com/LSSTDESC/baryon-challenge/blob/main/src/desc_baryon_challenge/likelihoods/firecrown_likelihood.py

PR #341, opened 2023-10-30
New load_likelihood_from_module

vitenti:
This PR introduces a new loading method to the system, which permits the direct loading of likelihoods from individual modules. This enables multiple likelihoods to be organized within a single Python package.


vitenti:
@tilmantroester does this PR address #340 ?

tilmantroester:
Yes, I think so.

Being able to specify the entry point (i.e., a function that's not necessarily named `build_likelihood`) would be a useful feature, so that different likelihood factories can be defined in the same file/module. But that's probably best done in a separate PR.

PR #345, opened 2023-11-02
Separating the cluster theory and models from Firecrown into an independent module.

mattkwiecien:
### Summary 
Changeset to separate the cluster theory and models logic from the internal workings of Firecrown.  As an added bonus, also updated the binned cluster number counts likelihood statistic to use `ModelingTools` in the expected `firecrown` workflow.

### Cluster module
* Added an `AbundanceData` class to wrap the sacc file and to be used by the likelihood 
  * The sacc file data needs to be manipulated to extract the relevant cluster number counts (and masses) for the likelihood
  * Put this code within this wrapper class to keep the likelihood more clear from this sacc-specific logic
* Removed firecrown infrastructure code and integration code from the `ClusterAbundance` class.   
* Added a new `Kernel` class to represent any ingredients in the cluster likelihood
  * Implemented kernels to represent the probability distribution P(z_spec|z_true), P(lambda|mass), etc
  * Created a `MassRichnessGaussian` of `Kernel`, anticipating having many of these.  This could potentially fall into yagni territory (you aren't gonna need it).
* Added a new `Integrator` class for the cluster module
  * Each integrator passes arguments to the cluster integrated differently, so we added this abstraction to allow specific integrators wrap and call the cluster abundance with their custom mapping
  * Implemented a NumCosmo and SciPy integrator as an example.

### Cluster likelihood factory function
Modified the cluster likelihood to use dependency injection for all of the choices for our likelihood.  Or in other words, moved all of the specific choices (integrator, cluster kernel ingredients, data vector) to the top of the call stack, so the user can swap in and out components to the likelihood.

### Other firecrown changes
* Added `ClusterAbundance` to `ModelingTools`.  
  * As a part of this, I have `ParamsMap` passing through `prepare`.
  * This may need to change now that `ModelingTools` is an updatable...?
* Added to `setup.cfg` instructions to tell mypy to ignore tests.  Not sure if we want to keep this.

### Tests
* Upped test coverage of all cluster code to 100%
* Ensured previous tests and new tests are all still passing
* Total runtime of all tests is < .5 seconds.  All tests except two take < 0.01 seconds.  Those two tests are almost regression, maybe we want to flag them as such?


firecrown/likelihood/gauss_family/statistic/binned_cluster_number_counts.py
```diff
@@ -0,0 +1,149 @@
...
+        if self.use_cluster_counts:
+            theory_vector_list += cluster_counts
+
+        if self.use_mean_log_mass:
+            theory_vector_list += self.get_binned_cluster_masses(tools, cluster_counts)
+
+        return TheoryVector.from_list(theory_vector_list)
+
+    def get_binned_cluster_masses(
+        self, tools: ModelingTools, cluster_counts: List[float]
```
marcpaterno, line 107: We propose passing the same enumeration flag type introduced above to choose which integral to do. If the passed value is not consistent with the state of the object, raise an exception. If more than one integral is requested, raise an exception.

Consider, rather than passing the enum value indicting which integral to do, instead just iterating of the enumeration data member.

tests/test_cluster_kernels.py
```diff
@@ -0,0 +1,224 @@
...
+
+    mass_proxy_limits = (1.0, 10.0)
+    z_proxy_limits = (0.1, 1.0)
+
+    truth = np.array(
+        [
+            0.77657274,
+            0.96966127,
+            0.99286409,
+            0.99780586,
```
mattkwiecien, line 109: Regression test

mattkwiecien (commented):
Addition changes requested

marcpaterno (approved):

Issue #346, opened 2023-11-09
Save predictions to sacc


tilmantroester:
A common use case is producing model predictions and then using those prediction as mock data for some downstream tasks. For that the model predictions need to be saved to a sacc object, for which the functionality does not yet exist in firecrown.

There are multiple implementations of this floating around in DESC (e.g. TXPipe, augur, as well as a number of notebooks), standardising one implementation in firecrown would reduce the overall maintenance burden across the various projects that rely on this functionality.

vitenti:
If we understood correctly your request and your PR, the additional information that would be necessary to use a computed theory vector to update a SACC object would be the list of indices. In this case we can save the SACC indices inside likelihood and have a method to return it. Thus, we can create a separate function that take SACC, theory vector and indices as input and update the SACC object. What do you think about this solution, since updating the SACC object is not part of the likelihood workflow we think that a separated function would be easier to test. We have however a question:

- In the PR you saved the last computed theory vector to use in the SACC update, is there a reason for that (will it take place during a full MCMC)? Otherwise, it is simpler to keep these separated, otherwise we need to make sure that the cached quantities are invalidated at the correct time.

tilmantroester:
This could be done in a separate function as well. My reason for putting it in `Likelihood` is that this allows access to the setup of the statistics and their internal state. For example to allow more sophisticated checking of which parts of the data vector are being saved.
I don't think assuming a specific workflow is a good approach here. Different users have different workflows. For example, the current design around a sampling workflow is ill-suited for a modelling workflow, which is already causing problems.

As for the issue with the caching: firecrown already caches the predictions (at least for two-point stats) and uses them downstream, `get_theory_vector` just gives a proper API for this. This is useful whenever you want to do anything beyond computing the chi2, such as plotting or saving the predictions.
firecrown has a very stateful design, so this doesn't seem out of place to me.

tilmantroester:
This has been addressed with #390 and #349.

Issue #323, opened 2023-10-10
Add support for posterior predictive distributions


tilmantroester:
This requires functionality to draw samples of data vectors from the likelihood and passing them back to the sampling framework.

The ability to draw samples from the likelihood is useful in other contexts as well, such as generating mock data vectors.

vitenti:
This is already supported by NumCosmo's connector. Moreover, in Augur you can find code to do that, see for example [srd_y1_3x2_like.py](https://github.com/LSSTDESC/augur/blob/master/examples/srd_y1_3x2_like.py), where they generate a data vector from a theory vector and build a likelihood to be used by any framework.

marcpaterno:
@tilmantroester is there something that Augur does, or something part of what it does, that you thing should be moved from Augur to Firecrown?

tilmantroester:
There are two reasons why I think this should be in firecrown:
One reason is that it's easiest to create the PPD draws while sampling instead of trying to create them after the fact. For a Gaussian likelihood with fixed covariance doing it in a post processing step is relatively straightforward if the model predictions get saved during sampling but for other likelihoods this might require re-evaluating the likelihood at a large number of points, which we want to avoid.
Drawing posterior predictive samples conditioned on parts of the data vector is probably easier to do in firecrown as well, since the description on how the data vector is structured is readily available there.
The other reason is that I might want to be able to use firecrown and generate mock data without the augur dependency, especially when building experimental pipelines.

joezuntz:
The ability to return data vectors is also useful for general debugging, and I'd recommend saving the information to do this.

However, in cosmosis I did find that the one case where this was slow compared to likelihood evaluations was Supernovae, so perhaps make it optional?

vitenti:
The CosmoSIS connector presently includes a section in the DataBlock labeled data_vector. This section contains three elements: `firecrown_theory` (the theory vector), `firecrown_data` (the data vector), and `firecrown_inverse_covariance` (the inverse covariance). To have these components written in the output chains, you can add them to the CosmoSIS `.ini` file under the `extra_output` section.

This behavior is automatically enabled for GaussFamily likelihoods, but the current implementation is not considered ideal. We are working to refine this process, with the goal of achieving the same outcome using `DerivedParameter`. The reason for the delay in implementing this change is the inherent difficulty of handling vector-derived parameters without resorting to the solution of appending `_n` to the derived parameter name to match the vector index.

[...]

PR #349, opened 2023-12-06
Save predictions to sacc

tilmantroester:
This PR addresses #346.

It writes the predicted theory vector into a sacc object, replacing the corresponding data point. By default the method raises an error if the sacc object contains data points that are not being overwritten by the predictions to avoid inconsistency.

Future additions to the functionality would be to draw a sample from the likelihood, instead of using the (zero-noise) theory prediction.

One discussion point is how to store the predictions. At the moment, only `TwoPoint` does this and uses `predicted_statistic_` for this. I've used `predicted_statistic_` for now for compatibility with existing code but I wouldn't mind renaming this to `theory_vector` to match the `data_vector`.

The test doesn't cover the checking of the prediction against the content of the sacc object yet.


tilmantroester:
Apart from the missing test and mypy errors, this can be reviewed I think @marcpaterno @vitenti

vitenti:
@tilmantroester , please, take a look on the modification to see if you agree with the current version.

firecrown/likelihood/gauss_family/gauss_family.py
```diff
@@ -98,28 +115,91 @@ def read(self, sacc_data: sacc.Sacc) -> None:
...
+            if set(sacc_indices.tolist()) != set(sacc_data.indices()):
+                raise RuntimeError(
+                    "The predicted data does not cover all the data in the "
+                    "sacc object. To write only the calculated predictions, "
+                    "set strict=False."
+                )
+
+        # Adding Gaussian noise defined by the covariance matrix.
+        assert self.cholesky is not None
+        predictions += np.dot(self.cholesky, np.random.randn(len(predictions)))
```
tilmantroester, line 180: Adding noise should be optional. A common use case is that we want a noise-free theory prediction as the data vector to test our models and pipelines.

tilmantroester:
Thanks for the help! I think adding noise to the prediction should be factored out of the functionality of saving the prediction. That way the functionality can also be used to save predictions+noise during sampling and would address #323.

vitenti:
@tilmantroester , what about making it optional? Take a look on the last commits.

tilmantroester:
I'd move `make_realization` to `Likelihood` or `GaussFamily`. Other likelihoods should still be able to save theory predictions, even if the sampling from the likelihood isn't implemented yet. We could add an abstract method `Likelihood.sample`, with the Gaussian and Student-T likelihoods implementing the specifics (I have the math for the latter).

vitenti:
> I'd move `make_realization` to `Likelihood` or `GaussFamily`. Other likelihoods should still be able to save theory predictions, even if the sampling from the likelihood isn't implemented yet. We could add an abstract method `Likelihood.sample`, with the Gaussian and Student-T likelihoods implementing the specifics (I have the math for the latter).

I think that we came up with the same solution, I just implemented your suggestion. Please, take a look.

vitenti:
@tilmantroester I think that it is now complete. Please tell me if you think there is something else to do here, otherwise we will review and merge it during our next Firecrown meeting.

tilmantroester:
Looks good to me, thanks!

firecrown/likelihood/likelihood.py
```diff
@@ -83,15 +83,31 @@ def __init__(self, parameter_prefix: Optional[str] = None) -> None:
...
 
+    def make_realization_vector(self) -> npt.NDArray[np.float64]:
+        """Create a new realization of the model using the previously computed
+        theory vector and covariance matrix.
+        """
+        raise NotImplementedError(
+            "This class does not implement make_realization_vector."
+        )
+
+    def make_realization(
```
marcpaterno, line 98: `mypy` misses that this method does not return the required `sacc.Sacc` object.
This method should raise a `NotImplementedError`, so that it is only an error to use a subclass that lacks this method if that method is called.

firecrown/updatable.py
```diff
@@ -392,10 +406,11 @@ def append(self, item: Updatable) -> None:
             )
         super().append(item)
 
-    def __setitem__(self, key, value) -> None:
+    def __setitem__(self, key, value):
         """Set self[key] to value; raise TypeError if Value is not Updatable."""
         if not isinstance(value, Updatable):
             raise TypeError(
-                "Values inserted into an UpdatableCollection must be Updatable"
+                "Only updatable items can be appended to an UpdatableCollection"
```
marcpaterno, line 413: Consider changing 'updatable' -> 'Updatable'

tests/likelihood/gauss_family/test_const_gaussian.py
```diff
@@ -33,13 +40,161 @@ def test_get_cov_works_after_read(trivial_stats, sacc_data_for_trivial_stat):
...
+    likelihood = ConstGaussian(statistics=trivial_stats)
+    likelihood.read(sacc_data_for_trivial_stat)
+    likelihood.update(trivial_params)
+    with pytest.warns(DeprecationWarning):
+        data_vector, theory_vector = likelihood.compute(ModelingTools())
+    assert np.all(data_vector == np.array([1.0, 4.0, -3.0]))
+    assert np.all(theory_vector == np.array([1.0, 1.0, 1.0]))
+
+
+def test_chisquared_compute_vector_not_implemented(
```
marcpaterno, line 177: Consider renaming: test_compute_chisquared_works_if_compute_theory_vector_raises_not_implemented_error

tests/likelihood/gauss_family/test_const_gaussian.py
```diff
@@ -99,3 +254,123 @@ def test_after_read_all_statistics_are_ready(
...
+    sacc_data_for_trivial_stat: sacc.Sacc,
+    tools_with_vanilla_cosmology: ModelingTools,
+):
+    likelihood = ConstGaussian(statistics=trivial_stats)
+    likelihood.read(sacc_data_for_trivial_stat)
+    params = firecrown.parameters.ParamsMap(mean=10.5)
+    likelihood.update(params)
+    likelihood.compute_chisq(tools_with_vanilla_cosmology)
+
+    new_sacc = likelihood.make_realization(sacc_data_for_trivial_stat)
```
marcpaterno, line 270: Consider adding pytest-rerunfailures to our list of required Conda packages, and decorating this test with `@pytest.mark.flaky(reruns=2)`. This should reduce the random failure probability from 1 in a million to a more comfortable 1 in a trillion.

This decoration should probably be done to all the tests that are statistical in nature.

tests/likelihood/gauss_family/test_const_gaussian.py
```diff
@@ -99,3 +254,123 @@ def test_after_read_all_statistics_are_ready(
...
+    # minus the covariance of the theory vectors.
+    covariance = np.cov(np.array(data_vector_list).T)
+    assert_allclose(
+        (covariance.diagonal() - var_exact) / np.sqrt(2.0 * var_exact**2 / 999.0),
+        1.0,
+        atol=5,
+    )
+
+
+def test_make_realization_no_noise(
```
marcpaterno, line 358: This test *should not be decorated* for re-running failures; it is not a statistical test.

marcpaterno (commented):
The Firecrown team will make the changes we have requsted.

firecrown/likelihood/gauss_family/gauss_family.py
```diff
@@ -98,28 +115,50 @@ def read(self, sacc_data: sacc.Sacc) -> None:
...
         self.cov = cov
         self.cholesky = scipy.linalg.cholesky(self.cov, lower=True)
         self.inv_cov = np.linalg.inv(cov)
 
         self.state = State.READY
 
     @final
-    def get_cov(self) -> npt.NDArray[np.float64]:
-        """Gets the current covariance matrix."""
+    def get_cov(self, statistic: Optional[Statistic] = None) -> npt.NDArray[np.float64]:
```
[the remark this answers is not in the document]
tilmantroester: The application I wrote this for was to get error bars when plotting the data vector. The idea was specifically to abstract away the indices and instead use the statistics, since that's what the user interacts with. I see the use of passing a list of statistics though, to get their corresponding sub-matrix.

firecrown/likelihood/likelihood.py
```diff
@@ -83,15 +83,31 @@ def __init__(self, parameter_prefix: Optional[str] = None) -> None:
...
-        self.predicted_data_vector: Optional[npt.NDArray[np.double]] = None
-        self.measured_data_vector: Optional[npt.NDArray[np.double]] = None
-        self.inv_cov: Optional[npt.NDArray[np.double]] = None
-        self.statistics: UpdatableCollection = UpdatableCollection()
-
     @abstractmethod
     def read(self, sacc_data: sacc.Sacc) -> None:
         """Read the covariance matrix for this likelihood from the SACC file."""
 
+    def make_realization_vector(self) -> npt.NDArray[np.float64]:
```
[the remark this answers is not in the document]
tilmantroester: From a statistics point of view, the likelihood is the distribution of the data. So from that perspective, `Likelihood` should be able to compute the log probability (that's already there) as well as sample from the distribution (what's being implemented here).

firecrown/likelihood/gauss_family/gauss_family.py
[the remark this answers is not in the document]
marcpaterno: We will expand the interface to have a single method `get_cov` that can accept:

1. a single `Statistic`
2. a list of `Statistic`
3. a single `np.ndarray` (the indices)
4. a list of `np.ndarray` (a list of indices)

In the case when a stat or list of stats is passed in, we need also to make sure that the stat (or all the stats) that have been  passed are in the likelihood object on which we've called the `get_cov` method. We will make the code verify this.

We also have to specify the order in which entries appear in the returned matrix. We propose to respect the order of the entries of the list of statistics (or of numpy arrays), so that the user-specified order of the list controls the ordering of the elements in the returned matrix result, rather than the order of the entires in the SACC data object controlling the ordering of the entries in the returned matrix.

For example, if we pass a list of stats of length 3 (note the order of the entries in this passed to `get_cov`:
stats1 -> 0:9, stats2 -> 10:19, stats3 -> 20:29 => `get_cov([stats1,stats3,stats2]) `-> 0:9 + 20:29 + 10:19

vitenti (approved):

PR #353, opened 2024-01-08
Refactoring the integration / integrand workflow in the cluster count calculation

mattkwiecien:
This PR changes the cluster number counts likelihood workflow to support potentially more complex integrands.

* Removed the `Kernel` abstract class and concept of a shared interface between components of the integrand. 
  * The abstraction that this introduces doesn't fit well with the use case of kernels that may take an arbitrary number of arguments.
  * This includes removing the automated integrand building code from `ClusterAbundance`.  The responsibility of assembling the integrand for the cluster number counts is now placed on the user in the form of `ClusterRecipe`
* Added `cluster_recipe` module.  This is where users can assemble a cluster number counts integrand using the pieces in firecrown.models.cluster
  * Updated the statistic to accept a cluster recipe. 
  * Each cluster recipe defines (1) the statically typed method which computes a prediction for the number of clusters in redshift, richness bin and (2) the mapping from the numerical integrator to the statically typed integrand.
* Added a `pylint_plugins` top level directory that enables ignoring of specific warnings on specific classes.  Included an example to ignore the duplicate code warning (we agreed this should not apply to `ClusterRecipe`).
*  Removed old unused code
* Added new tests.
* Disabled the duplicate-code warning in tests, and too-few-public-methods in firecrown.

### DIFF
--- a/firecrown/models/cluster/recipes/murata_binned_spec_z.py
+++ b/firecrown/models/cluster/recipes/murata_binned_spec_z.py
@@ -13,7 +13,6 @@ from firecrown.models.cluster.properties import ClusterProperty
 from firecrown.models.cluster.recipes.cluster_recipe import ClusterRecipe
 
 
-# pylint: disable=R0801
 class MurataBinnedSpecZRecipe(ClusterRecipe):
     """Cluster recipe using the Murata 2019 binned mass-richness relation and assuming
     perfectly measured spec-zs."""
--- a/firecrown/models/cluster/recipes/murata_binned_spec_z.py
+++ b/firecrown/models/cluster/recipes/murata_binned_spec_z.py
@@ -1,3 +1,4 @@
+"""Module for defining the classes used in the MurataBinnedSpecZ cluster recipe."""
 from typing import Callable, Optional, Tuple
 
 import numpy as np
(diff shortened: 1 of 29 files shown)


marcpaterno:
We agreed to have the `tests/pylintrc` configuration remove the check for duplicated code (but only in the tests).

firecrown/models/cluster/recipes/murata_binned_spec_z.py
```diff
@@ -12,7 +13,11 @@
 from firecrown.models.cluster.recipes.cluster_recipe import ClusterRecipe
 
 
+# pylint: disable=R0801
```
marcpaterno, line 16: We discussed the disabling of the duplicated code complaint and agree that it should be relaxed for recipes.
We'd prefer to see the disabling being done in a file or directory level through `pylintrc` to having it on each class.

We should also use the named form, rather than the numeric form, of what is being disabled.

mattkwiecien: This is surprisingly hard to do with pylint.  The best solution I've found is that you can write custom pylint plugins and dynamically append ignore statements to specific classes.  

I implemented this by adding a new `pylint_plugins` directory to the root of the project (and added a readme explaining why).  I then added a `duplicate_code.py` which contains the custom plugin that can ignore duplicate_code, and inside that I only add the ignore comment to `firecrown.models.cluster.recipes` classes.

Then, I updated the pylint rc files to load this plugin by default.  

I pushed this up in the most recent commit.

marcpaterno (approved):

PR #358, opened 2024-01-22
Test PR for Codecov coverage tool 

mattkwiecien:
I'm using this PR to test and customize the status checks from Codecov.  

Codecov reports are being uploaded to codecov.io already as a part of this branch.  Now we want to customize the differential coverage to the rules we want for firecrown.  I am using this PR to do adjust these settings.


mattkwiecien:
Closing - codecov needs a single initial base commit to compare against before it can start doing differential coverage checks.  To test this I will be using a branch and a branch off of that branch to configure the tool.

Issue #375, opened 2024-02-10
Python 3.9+ no longer needs typing.List, Set, Dict, or Tuple


marcpaterno:
As of Python 3.9, the builtins list, set, dict and tuple can be used in type annotations, rather than the types in the typing module (which were needed for Python 3.8 and earlier). We should simplify our code by using the builtin types.

PR #376, opened 2024-02-10
375 python 39+ no longer needs typinglist set dict or tuple

marcpaterno:
This will close issue #375


mattkwiecien:
Now that we have codecov checks working ( #381 ) and coveralls is failing here, we may want to merge #381 first.  That way we could see coveralls fail and codecov succeed for this edge case of removing code decreasing total coverage.

vitenti (approved):

PR #393, opened 2024-02-22
Removing integration tests from coverage report.

### LINKED ISSUE #387: Exclude integration tests from code coverage
Moving the examples into integration tests to be run by pytest artificially inflate our code coverage metrics (e.g. `TwoPoint` has much higher coverage now without adding any unit tests).  

To keep our code coverage meaningful, we should remove added coverage due to these examples from our coverage reports.

mattkwiecien:
Closes #387


marcpaterno (approved):
The code coverage is going down because we are no longer running examples and pretending they are tests.

PR #397, opened 2024-03-01
Proposal for likelihood specification and factory API

tilmantroester:
This is a more-or-less working prototype of a yaml-based likelihood specification, following the [discussion thread](https://github.com/LSSTDESC/firecrown/discussions/308#discussioncomment-7589995).

Some features I think are nice:
- Separation of model and data specification: specifying the model does not require knowledge about the data
- Flexible: by using jsonargparse for dependency injection, the model specification can use systematics directly from firecrown or elsewhere, without the factory having to know about these systematics.
- I tried to separate the yaml-parsing logic from the likelihood factory logic, so the yaml format is not a hard requirement

Some obvious things that are still missing
- Clusters and SN. This only supports 2pt at the moment. I'm not very familiar with clusters workflow and an earlier version I had has stopped working.
- Features such as scale and bin cuts, but that should be straightforward to implement.
- Robustness, testing, etc.

I hope this PR can function as a starting point of a discussion on where this specification might not be up to our tasks, potential implementation pitfalls, and other suggestions and comments.

### DIFF
--- a/examples/des_y1_3x2pt/factory_example.py
+++ b/examples/des_y1_3x2pt/factory_example.py
@@ -1,0 +1,125 @@
+import numpy as np
+
+import matplotlib.pyplot as plt
+
+import sacc
+
+import pyccl as ccl
+
+from firecrown.likelihood.likelihood import NamedParameters
+from firecrown.parameters import ParamsMap
+
+from firecrown.factory.factory import build_likelihood
+
+
+config_str = """
+two-point:
+    weak_lensing:
+        global_systematics:
+            IA:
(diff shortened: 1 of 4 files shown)


vitenti:
Hi @tilmantroester , we have been working on the redesign of two point in the last two meetings and we are using your work here as a source of ideas to the new design. Soon we will have a separate PR branch with our code to discuss.

marcpaterno:
Tilman, does the YAML configuration facility that we have started to add in release 1.8 meet this need? We recognize that not everything is configurable this way (yet). But we think that everything mentioned in this PR is now configurable.

If you agree, please close this issue.

vitenti:
It seems that all proposed features are already implemented. Please open another issue if you find there is something missing.

PR #400, opened 2024-03-20
Integrate halo model for intrinsic alignments in 3x2-point systematics.

chrgeorgiou:
This PR integrates the halo model for intrinsic alignments as a potential systematic for 3x2-point likelihoods. It adds a new ```WeakLensingSystematic``` class called ```HMAlignmentSystematic``` and the appropriate behaviour for when a tracer has this systematic attached to it. It also extends the ```ModelingTools``` for computing halo model-related quantities. Finally, it computes the correct power spectrum given the input tracer.


firecrown/modeling_tools.py
```diff
@@ -25,13 +25,21 @@ def __init__(
         pt_calculator: Optional[pyccl.nl_pt.EulerianPTCalculator] = None,
         pk_modifiers: Optional[Collection[PowerspectrumModifier]] = None,
         cluster_abundance: Optional[ClusterAbundance] = None,
+        hm_definition: Optional[pyccl.halos.MassDef] = None,
```
tilmantroester, line 28: I think passing the HMCalculator object directly (or a wrapper around it), similar to the `PTCalculator`, would make the API more future proof and also abstract away these CCL details.

chrgeorgiou:
Hi @marcpaterno and @vitenti, given the new update of Firecrown with factories, I would like to update this PR and try to finalise it. Is this a lot of work, and would I need to do this by myself?

marcpaterno:
@chrgeorgiou, please go ahead and finish your commits that will  solve the conflicts. Please also run the pre-commit checks, which will let you locally see the same results that the GitHub CI will produce. You can do this with
```
bash pre-commit-check
```
If you are struggling too much to get the checks passed, please go ahead and push the commits anyway, and we may be able to help solve some outstanding issues.

marcpaterno:
@chrgeorgiou I have fixed as many of the CI failures as I know how to fix. The remaining items that need to be fixed before this PR can be merged are the following.

1. There are several 'FIXME' annotations, which pylint flags as failures. Please only use FIXME for reminders to yourself of what needs fixing before a PR is finished. If you want a marker for code that is a candidate for later enhancement, please use TODO.
2. All newly or modified code needs to be tested. The test coverage output from the CI will show you what lines of code are untested. If you have installed the development version of Firecrown through the recommended method, you can run the test coverage locally with the command:
```
 python -m pytest -vv --runslow --cov firecrown --cov-report html --cov-branch
```
Then open the file `htmlcov/index.html` with a web browser to look at the results.

chrgeorgiou:
Hi Marc, thanks so much for this! I was still working on some of the development for this PR (based on the comment of Tilman above) but it is now finished. I have removed the "FIXME" statements, indeed they were simply for me personally to not forget to address. 

I have run the pre-commit-check and it seems okay. Let me know if I need to develop any tests for the new code.

firecrown/models/two_point.py
```diff
@@ -59,26 +59,120 @@ def calculate_pk(
...
+    )
+    if not (tracer0.has_hm and tracer1.has_hm):
+        IA_bias_exponent = (
+            1  # IA bias if not both tracers are HM (doing GI correlation).
+        )
+        if ("galaxies" in tracer0.tracer_name) or ("galaxies" in tracer1.tracer_name):
+            other_profile = pyccl.halos.HaloProfileHOD(
+                mass_def=hm_calculator.mass_def, concentration=cM_relation
+            )
+            other_profile.ia_a_2h = -2  # tracer0.bias
```
chrgeorgiou, line 98: @marcpaterno I have a question here: Here we will have one tracer with `has_hm` true and another tracer with `"galaxies" in tracer.tracer_name`. This means correlations between galaxy position and galaxy shapes where the shape statistics are modelled with the halo model for intrinsic alignments. 

To compute this correlation, we need the galaxies tracer to have a linear galaxy bias systematic, and we need to querry the value of this bias. I haven't been able to find a way to get that value from the tracer object in this part of the code. Is there a way to do this?

marcpaterno: My understanding of the original design was for systematics to be independent, and that the way to make sure a consistent set of systematics was constructed was to have a factory create them from the appropriate configuration. In addition, systematics are always optional -- they are used to adjust the calculations done by models, but they aren't required parts of those models. This seems like an addition that requires a specific systematic be in place for the two-point statistic to do its job at all. @chrgeorgiou is that correct?

@vitenti can you comment on this?

tilmantroester: I think this is not so much about there being a linear galaxy bias systematic but there being a well-defined (linear) galaxy bias for the tracers. 

If it's a number count tracer that's neither PT or HM, then it should be a plain ccl `NumberCountsTracer` (https://github.com/LSSTDESC/firecrown/blob/master/firecrown/likelihood/number_counts.py#L474) and should have a well-defined bias value. 

For PT and HM number counts, this gets more complicated. In case of PT, you could pull out `b1` but this won't capture the full scale-dependent galaxy bias.

Weak lensing tracers don't have galaxy bias at all.

My guess that this needs more logic to handle the different combinations of tracers that can occur.

firecrown/modeling_tools.py
```diff
@@ -29,13 +29,17 @@ def __init__(
...
+        cM_relation: None | str = None,
         pk_modifiers: None | Collection[PowerspectrumModifier] = None,
         cluster_abundance: None | ClusterAbundance = None,
         ccl_factory: None | CCLFactory = None,
     ):
         super().__init__()
         self.ccl_cosmo: None | pyccl.Cosmology = None
         self.pt_calculator: None | pyccl.nl_pt.EulerianPTCalculator = pt_calculator
+        self.hm_calculator: None | pyccl.halos.HMCalculator = hm_calculator
+        self.cM_relation: None | str = cM_relation
```
tilmantroester, line 42: Why is the c-M relation stored here? Can't this be accessed through the `HMCalculator`? If not, it'd probably be cleaner to change `HMCalculator` such that it is accessible in order to avoid inconsistent states in firecrown. And avoid the need to carry along other things like the mass definition in the future.

firecrown/models/two_point.py
chrgeorgiou: Hi Marc, yes your understanding is correct. When cross-correlating galaxies (number counts tracers) with galaxy shapes using this model, the power spectrum is going to depend on the assumed galaxy bias model attached to the number counts tracer. The implementation I have here is only valid for a linear galaxy bias model. Implementing more complicated models is an open question and out of the scope here. 

@tilmantroester I thought that for number counts tracers with non-linear galaxy bias the 2-halo terms for the halo model would have to be carefully recomputed to inlucde the contributions from higher-order galaxy bias terms. So a `b1` would only be an approximation, but maybe I am mis-understanding this.

I guess the broader question here is whether we want to support this feature in Firecrown, having the option for different systematics to "interract" - which can happen for these non-linear systematics models like the halo model. Perhaps it's a good idea to have a discussion on this first before scoping more Firecrown development.

For this PR I can leave out the support for number counts - weak lensing tracers (galaxy-galaxy lensing) which would make the halo model for IA usable only for cosmic shear at the moment. And decide later what are the correct steps forward for this (for example, we might want this to also be used for galaxy clustering measurements, which now is not supported, or for stellar mass function predictions that now Firecrown does not cover at all). So, how about I raise an error when such a correlation is called for now?

firecrown/likelihood/weak_lensing.py
```diff
@@ -343,6 +393,29 @@ def create_tracers(self, tools: ModelingTools):
...
+            ccl_wl_dummy_tracer = pyccl.WeakLensingTracer(
+                ccl_cosmo,
+                has_shear=False,
+                use_A_ia=False,
+                dndz=(tracer_args.z, tracer_args.dndz),
+                ia_bias=(tracer_args.z, np.ones_like(tracer_args.z)),
+            )
+            ia_tracer = Tracer(
+                ccl_wl_dummy_tracer,
+                tracer_name="intrinsic_hm",
```
tilmantroester, line 411: Maybe call it `intrinsic_alignment_hm` for clarity?

firecrown/models/two_point.py
chrgeorgiou: Right now, the implementation of the model is not written down for number counts vs shape tracers in any article. The most straightforward thing to do would be to have the number counts tracer also following a halo model but that's not been done in data before (for both clustering and IA/Lensing) and is not scoped in this PR.

firecrown/models/two_point.py
```diff
@@ -59,26 +59,120 @@ def calculate_pk(
...
+    pk_1h = pyccl.halos.halomod_Pk2D(
+        cosmo=ccl_cosmo,
+        hmc=hm_calculator,
+        prof=profile0,
+        prof2=profile1,
+        a_arr=a_arr,
+        get_2h=False,
+    )
+    # Compute here the 2-halo power spectrum
+    C1rhocrit = (
```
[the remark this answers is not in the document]
chrgeorgiou: Yes, I agree. I think for this PR a quick solution is as you say to check the tracers and throw an error. But it is a broader question how we want this to be implemented in Firecrown and how much effort should be put into this.

firecrown/modeling_tools.py
chrgeorgiou: The halo model calculator does not take a concentration mass relation as an argument, it only accepts mass definition, halo mass/bias functions and some extra parameters related to the integration. The concentration is needed to initialise the IA halo model profile (and other HOD profiles). 

Mass definition, halo mass and bias functions are carried by the halo model calculator. Do you think this is okay then as an implementation then for halo model-based tracers?

tilmantroester: Right, I guess the c-M relation is more of a halo-profile-specific property, rather than a halo-model property. Would it make sense to have the c-M relation set has part of the systematic? That way different halos (e.g. for baryons) could have different profiles and concentrations. It would also not increase the API surface of `ModellingTools`. What do you think?

marcpaterno:
The new instance variable in `ModelingTools`, `cM_relation`, has type `str|None`. But the accessor `get_cM_relation` returns type `pyccl.halos.Concentration`. This is a base class with many derived types and not a `str`. I do not understand why `mypy` did not spot this problem, but I don't understand how this is used.

The only test I found that uses this method it just verifies that if called when the instance variable is `None` that an exception is raised. I do not see anything that uses the return value of the method.

Can you explain how this is intended to work?

firecrown/modeling_tools.py
chrgeorgiou: Yes, it would be better to connect this to the tracer but I could not find a way to have a tracer-specific property that was not `Updatable` and was also a `str`. @marcpaterno Is there a way to do this? For example adding in `__init__` of the `HMAlignmentSystematic` something like
```
self.cM_relation = parameters.register_new_fixed_parameter(
            default_value='Duffy08'
        )
```

chrgeorgiou:
> The new instance variable in `ModelingTools`, `cM_relation`, has type `str|None`. But the accessor `get_cM_relation` returns type `pyccl.halos.Concentration`. This is a base class with many derived types and not a `str`. I do not understand why `mypy` did not spot this problem, but I don't understand how this is used.
> 
> The only test I found that uses this method it just verifies that if called when the instance variable is `None` that an exception is raised. I do not see anything that uses the return value of the method.
> 
> Can you explain how this is intended to work?

Ah, thanks for catching this, it was a left-over from previous development. Actually, the `cM_relation` can be `str` or `pyccl.halos.Concentrartion` object, the systematic will be used correctly with either. If we can move the variable `cM_relation` from modelling tools to be under the `HMAlignmentSystematic`, this will then come from the CosmoSIS ini file (or the equivalent of Cobaya) and should be just a `str`. I will modify the code now to have this specified as a `str`.

firecrown/modeling_tools.py
tilmantroester: I'd try adding something like an `ia_hm_cm` field to `WeakLensingArgs`, which gets populated by `HMAlignmentSystematic`. Since this is not a parameter that gets sampled, it can just be an attribute of `HMAlignmentSystematic`. I don't know how such configuration arguments are passed through the various factories though.

firecrown/modeling_tools.py
```diff
@@ -29,13 +29,17 @@ def __init__(
         self,
         *,
         pt_calculator: None | pyccl.nl_pt.EulerianPTCalculator = None,
+        hm_calculator: None | pyccl.halos.HMCalculator = None,
+        cM_relation: None | str = None,
```
marcpaterno, line 33: We would prefer for the cM_relation to be the concrete `Concentration` object, rather than a string which will eventually be converted to the right object.

firecrown/modeling_tools.py
marcpaterno: We want to follow up on improving how this is done, but will merge this PR as-is since it does make a step toward the solution.

marcpaterno (approved):
We will merge this because it takes a good step forward. After the merge we will take into consideration the discussion in the comments for this PR, to determine what in the systematics handling needs to be improved to make this task easier.

firecrown/likelihood/weak_lensing.py
```diff
@@ -343,6 +393,29 @@ def create_tracers(self, tools: ModelingTools):
...
+                has_shear=False,
+                use_A_ia=False,
+                dndz=(tracer_args.z, tracer_args.dndz),
+                ia_bias=(tracer_args.z, np.ones_like(tracer_args.z)),
+            )
+            ia_tracer = Tracer(
+                ccl_wl_dummy_tracer,
+                tracer_name="intrinsic_alignment_hm",
+                halo_profile=halo_profile,
+            )
```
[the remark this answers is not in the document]
chrgeorgiou: Yes, that is correct, I could not find a better way to do this so this was a quick workaround. I think it would be useful to have the Tracer objects carry with them the value of the systematic parameters, it would allow for flexibility when computing power spectra with more complicated models (e.g. non-linear galaxy bias, non-linear intrinsic alignment bias, the interaction between the two and their interaction with magnification).

firecrown/modeling_tools.py
chrgeorgiou: It can also be a `Concentration` object but then it has to be defined at the likelihood building level and I am not sure how it would be moved under the `WeakLensingTracer` object in this case. But I can imagine it's possible.

PR #405, opened 2024-04-02
Excerpt from twopoint_metadata containing two_point refactoring only.

vitenti:


vitenti:
This pull request was open to create a bugfix release. It was used to check the build and linting. It will not be merged.

PR #409, opened 2024-04-12
408 following our planned roadmap we should cease support for python 39

### LINKED ISSUE #408: Following our planned roadmap, we should cease support for Python 3.9
Our plan has been to stop support for a given Python version when numpy drops support for it. Numpy has (as of April 5, 2024) dropped support for Python 3.9

We should drop support for Python 3.9. and then make the various upgrades we could not do under Python 3.9:

1. replacing Union[X,Y] by X | Y in type declarations
2. replacing Optional[X] by None | X in type declarations
3. using kw_only=True in dataclasses

marcpaterno:
This will close #408

### DIFF
--- a/examples/des_y1_3x2pt/des_y1_3x2pt.py
+++ b/examples/des_y1_3x2pt/des_y1_3x2pt.py
@@ -1,7 +1,6 @@
 """The DES Y1 3x2pt likelihood factory module."""
 
 import os
-from typing import Union
 import sacc
 
 import firecrown.likelihood.gauss_family.statistic.source.weak_lensing as wl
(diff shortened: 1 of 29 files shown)


.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu", "macos"]
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.10", "3.11"]
```
vitenti, line 23: Why not adding 3.12 here?

vitenti (commented):
Should we remove coveralls from the CI?

.github/workflows/ci.yml
marcpaterno: I didn't think it was good to add 3.12 to our workflows when we would still get an installation failure because of the unavailability of tensorflow. When that is solved I agree we should add it.

vitenti (approved):

PR #444, opened 2024-07-19
Issue/439/cluster wl from mass

yyzhang:
Adding a new cluster model to output cluster delta sigma for given mass, redshift and radii. 

- Goal is that the new model class can be used as a function in cluster recipes. 
- Implemented unit test code.


vitenti:
Hello @yyzhang , we recently merged PR #461 that includes support for weak-lensing proxies through Delta Sigma. Please, tell us if that inclusion solves the problem you were trying to solve here. If so, please close this PR.

marcpaterno:
Hi @yyzhang , we are checking in again to see if the merged PR #461 hat includes support for weak-lensing proxies through Delta Sigma. Please, tell us if that inclusion solves the problem you were trying to solve here. If so, please close this PR.

PR #472, opened 2024-11-25
Relax relative tolerance from 1e-8 to 1e-6

marcpaterno:
## Description

The CI testing is currently failing because of 3 tests which require a numerical answer to match with a tolerance of 1e-8.
This PR will "fix" the problem by relaxing the tolerance to 1e-6.

## Type of change

Please delete the bullet items below that do not apply to this pull request.

* Bug fix (non-breaking change which fixes an issue)
## Checklist:

- [ ] I have 100% test coverage for my changes (please check this after the CI system has verified the coverage)

### DIFF
--- a/tests/test_bug398.py
+++ b/tests/test_bug398.py
@@ -150,7 +150,7 @@ def test_eval_cl_window_src2_src2() -> None:
     src2_src2.update(params)
 
     theory_vector = src2_src2.compute_theory_vector(tools)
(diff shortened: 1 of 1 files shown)


marcpaterno:
It was a change in CCL that caused the change in calculated results.

The purpose of these tests is to ensure no regressions in the theory calculations. Thus this is the wrong fix; we should update the expected values, and keep the tight tolerances.

PR #485, opened 2025-01-31
Modified selection function to be updatable

eduardojsbarroso:
## New feature requests and enhancements
I suggest we implement the completeness and purity kernels in firecrown/models/cluster/kernel.py with free parameters that can be sampled by cosmosis.

## Problem description
First of all, there is an error in the current kernel implementation. Current implementation is using logm/logmc whereas it should be m/mc, as per [Aguena, 2016](https://arxiv.org/abs/1611.05468). However I also propose that we implement these functions as we did in mass_proxy.py, where the parameters will be set by cosmosis instead of hard coded. Even if we do not sample over these parameters, they will be different for different input data (i.e. galaxy cluster catalogs obtained from different cluster finders). Also, I think it would be helpful if we can sample over the parameters. We can find a fit from cosmological simulations and then I suggest a gaussian prior centered on these fit parameters to give some freedom for the fit.

## Proposed solution
Change Completeness and Purity classes in firecrown/models/cluster/kernel.py to be updatable objects with 4 parameters each.


eduardojsbarroso:
Hellow @vitenti and @marcpaterno. I have made some updates to the completeness and purity code here. To use these functions on the cluster recipes, I need to create yet another cluster recipe file which will have a bunch of repeated code that `pylint` will give warnings again. Adding the purity kernel also involves one extra integral that we were skiping in the other cases. So I will not create this new recipe for now and later on, either I make another pull request to refactor all the recipes or we decide if we are moving on to something else.

marcpaterno (approved):

PR #494, opened 2025-03-17
431 Shifts and Stretch

### LINKED ISSUE #431: Photo-z stretch systematic
For the DESC SRD update, we need to be able to include a photo-z stretch parameter that changes the width of the photo-z distribution in addition to the photo-z shift.

JaimeRZP:
## Description

Bug fix in Shift and Stretch model to close #431 

Fixes # (issue)
#431 

## Type of change

Bug fix: Shift&Stretch was missing shift


vitenti:
Hi @JaimeRZP,  

While addressing the PR, I took the opportunity to clean up and improve certain parts of the code. During the process, I realized that the bug wasn’t actually in the section we initially suspected. Specifically, since `SourceGalaxyPhotoZShiftandStretch` subclasses `SourceGalaxyPhotoZShift`, calling `super().apply()` executes the `apply` method from `SourceGalaxyPhotoZShift` first, followed by the stretching operation. This suggests that the issue might lie elsewhere.  

It would be helpful if you could point out any specific code or scenarios that produce incorrect results so we can investigate further. That said, the PR includes several meaningful improvements, so I believe it’s worth merging if you’re comfortable with the changes I’ve made.

JaimeRZP:
Hi @vitenti, my bad then. I was familiar with an older version of the code and missed the subtlety you have described.
Thank you for taking over the PR and feel free to merge.

vitenti (approved):

Issue #531, opened 2025-07-03
Bug in bandpower window function logic


tilmantroester:
The core issue is that the `ells` that are returned by `sacc_data.get_ell_cl` are fundamentally different from what is in `bandpower_window.values`. The former is a description of the data point and what should be used for deciding scale cuts. The latter is a description for how the prediction for the data point should be calculated based on the window function. 
Both of these must be tracked separately, since they serve different purposes.

Note that the window function can have contribution from outside the cut scales and it is wrong to exclude part of the ell-range of the window function when computing predictions.

Suggestion: add a `window_ell` attribute to `TwoPointHarmonic`, which is used for the calculation for the theory prediction but not for the scale cuts (which are done on the `ells` from `sacc_data.get_ell_cl`).

The real space measurements will also need to have some form of window function support in the future.

vitenti:
@tilmantroester I initially replied on your PR, but now that I’ve seen the Issue, I agree with your assessment. Since the ell labels from `BandpowerWindow` serve distinct purposes, it makes sense to track both explicitly. Given that the labels/markers are important for defining the scale cuts, we can proceed with your proposed modification and add a window_ell attribute to TwoPointHarmonic. I can make this implementation but feel free to make it if you prefer.

PR #533, opened 2025-07-04
Test for bandpower bug

tilmantroester:
This adds a test which currently fails due to the bug described in #531


vitenti:
Hello @tilmantroester,
The current implementation in Firecrown applies scale cuts based on the window function: any observable that depends on an ell outside the specified range is removed. The test was failing because the window function extended far beyond its relevant domain (down to ~1.0e-300). I fixed this by clipping the window to remove irrelevant values, allowing the existing logic to correctly exclude the intended bins.

I believe the implementation you had in mind would associate each window function with a set of label/marker ell values (e.g., bin centers), and apply the cuts by comparing against these. If the current approach isn't sufficient and a label-based method is needed, we can easily implement it.

PR #539, opened 2025-08-02
Add warning for unused parameters in likelihood computations

vitenti:
## Description

This pull request adds a check for unused parameters in likelihood computations. If any keys are present in the input ParamsMap but not used during the likelihood evaluation, a UserWarning is issued. This behavior helps identify potential misconfigurations or typos in the parameter mapping.

A configuration flag was added to optionally raise an error instead of issuing a warning.

## Type of change

Please delete the bullet items below that do not apply to this pull request.

* Bug fix (non-breaking change which fixes an issue)

## Checklist:


tests/likelihood/test_likelihood.py
```diff
@@ -19,7 +19,7 @@ def compute_loglike(self, _: like.ModelingTools) -> float:
 class LikelihoodThatThrowsUnhandledError(like.Likelihood):
     """A likelihood that always throws a pyccl integration error exception."""
```
tilmantroester, line 20: Not directly connected to this PR but these tests confused me initially. Updating the doc string and maybe a comment on what's going on (I guess integration errors get ignored by setting the log like to -inf but other errors propagate?) would be useful I think.

firecrown/connector/cosmosis/likelihood.py
```diff
@@ -143,6 +143,8 @@ def execute(self, sample: cosmosis.datablock) -> int:
             # Cosmology will be generated by CCL. We want to be sure that CosmoSIS
             # has not also been configured to use CAMB.
             self.tools.prepare()
+
+        handle_unused_params(params=firecrown_params, raise_on_unused=False)
```
tilmantroester, line 147: I'd argue for defaulting to `raise_on_unused=True` and getting `raise_on_unused` from `config` to override by the user.

marcpaterno (changes requested):
We would like to see a few changes:

1. Change the default value of `raise_on_unused` to be `True`.
2. Make sure each of the connectors are tested (e.g. to ensure they don't have different behaviors when the value of `raise_on_unused` is specified).
3. Tests that the default value of the parameter is what is intended (this test can be done using the `inspect` module's `signature` function).

PR #576, opened 2025-10-29
Better updatables

vitenti:
## Description

Adds functionality to track parameter updates and generate detailed logs.
Introduces `assert_updatable_interface` to verify that the `Updatable` interface methods have not been overridden or modified.

## Type of change

## Checklist:


vitenti:
@marcpaterno

* I experimented with using a `rich.tree.Tree` object to produce the update logs. Do you think we should adopt that?
* I also added `assert_updatable_interface` to verify that the user-provided likelihood is a valid `Updatable`. Should we enforce this check during load?

marcpaterno (approved):

PR #603, opened 2026-01-09
Adding spectroscopic stretch (FoG model)

iagolops:
## Description

This commit implements a new systematic, `SpecZStretch`, which convolves the $N(z)$ distribution with a window chosen to model the Fingers-of-God effect. The theoretical approach for this modelling is discussed in [2008.08706](https://arxiv.org/abs/2008.08706). In this commit, we've added the Gaussian and the Lorentzian windows, where the systematic parameter is the velocity dispersion ($\sigma_v$) within the redshift bin, where its unity is $\text{km}/\text{s}$.

The (unnormalized) Gaussian convolution is given by:

$$
n_{conv}^{i}(z) = \int_{-\infty}^{\infty} n^i(s) \exp\left(-\frac{(z-s)^2}{2\sigma_{s,i}^2}\right) ds.
$$

Here, $\sigma_{s,i}$ is the redshift dispersion in the $i$-th redshift bin, which is related to the velocity dispersion by:

$$
\sigma_{s,i} = (1+ z_{mean,i})\frac{\sigma_{v,i}}{c}.
$$

The image below shows the result of this implementation. It shows the behavior of **cross-correlations** when applying this new systematic into spectroscopic $N(z)$ distributions.

cell_vs_sigma_v_rsd

We (I, @arthurmloureiro, and Bruno Moraes) implemented this systematic based on PhotoZShiftandStretch, where we added the two main functions `dndz_stretch_fog_gaussian` and `dndz_stretch_fog_lorentzian`, which are similar to `dndz_shift_and_stretch_active`.

**To-do**:

- Add tests
- Check how the user can change the kernel from Gaussian to Lorentzian in the `__init__` of `SourceGalaxySpecZStretch`

## Type of change

## Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have made corresponding changes to the documentation
- [ ] I have 100% test coverage for my changes (please check this after the CI system has verified the coverage)


arthurmloureiro:
> Hi @arthurmloureiro, we're sorry for the current messy state. As we have noted in previous CCL/Firecrown meetings, what you're observing is the ongoing process of our reorganization of the physical structure of the package. The messy `likelihood` module is the main thing that has _not_ yet been reorganized.

Hi @marcpaterno !

Thanks for letting us know that this is not the final stage of this refactoring. 

If I can voice a few concerns about the refactorings in Firecrown:
1) Please tell me that this labyrinth of imports is something temporary. I feel like there was a jump in orders of magnitude in the abstraction in firecrown. It's ok if this is is just a temporary structure while things are being changed, but if this is close to the final structure, it will be impossible for others to understand the code and develop for it. It's ever more abstract than it was before. I consider myself quite familiar with firecrown and I do not understand anything anymore.
2) Over the past few months, we have an increasing number of projects and pipelines that are dependent on firecrown. Although some more official pipelines are considered when structure changes, the other projects have a severe amount of overhead to keep up with the pace of changes in the import trees and function calls. This is causing several delays and impacting a lot of student's whose projects have very tight deadlines. 

About 2) my suggestion would be to not deprecate structure between small version changes or we should instead avoid having any student projects to use firecrown until a stable and immutable structure is chosen. First one would be best, given that there currently several student projects ongoing that are using firecrown.

PR #616, opened 2026-02-09
Generalize Tracer Framework and Add CMB Lensing Support

### LINKED ISSUE #606: `make_all_photoz_bin_combinations_with_cmb` has duplicated correlations (my bad!)
**Describe the bug**
Hi!

I noticed I pushed a bug into this metadata function that calculates the cross-correlation combinations with CMB Lensing by doing both cmb-kappa x galaxies and galaxies x cmb-kappa. I should have removed the second case but forgot:

https://github.com/LSSTDESC/firecrown/commit/0216ee3ea6ff8136c5435fccfaaedaca33cd9bd6#diff-b5687b19fe19ded8040c675fe0788a1f7b37b0bf9dfe0598e95c0a9cbd9137d5R514-R587 

I will push a PR to fix this.

**To Reproduce**
Use `make_all_photoz_bin_combinations_with_cmb` and see.

**Expected behavior**
We should have just one of the combinations (I,J) only instead of (I,J) and (J,I).

**Observed behavior**
(I,J) and (J,I).

**Output**
(I,J) only

**Configuration:**
Master branch and all other releases since CMB Lensing was implemented.

**Additional context**
I will fix this :)

### LINKED ISSUE #607: CMB Lensing Tracer Implemented as N(z) tracer post PR
**Describe the bug**
Hi Firecrown Devs,

While working with @xanderhenrik and testing the CMB lensing implementation, I noticed a serious issue in the current design: CMB lensing is being treated as an N(z) tracer, and this causes fundamental problems throughout the pipeline.

This behavior was introduced in this commit:
https://github.com/LSSTDESC/firecrown/pull/543/commits/4cfaa020a254385c8eb724460d252b2edf1db21e

Conceptually, assigning a redshift distribution to the CMB is incorrect. The CMB is a source plane at a fixed redshift, not a population with an N(z). Treating it as such leads to multiple downstream issues, particularly in the connectors (e.g. CosmoSIS). As far as I can tell, the connector logic extrapolates all redshift-dependent quantities up to the maximum redshift found across all N(z)s. When the CMB is included as an N(z) tracer, this effectively forces CosmoSIS/CCL to evaluate P(k) at the CMB redshift, which consistently breaks the calculation.

Previously, I implemented the CMB source as a distinct tracer type precisely to avoid this class of problems. Reintroducing it as an N(z) tracer reopens all of these issues and makes the CMB lensing support very fragile.

Related to this, the removal of the default CMB source redshift here:
https://github.com/LSSTDESC/firecrown/pull/543/commits/c35c0741740debe56f7fd413850b92a3c67456d7

also contributes to the problem. While configurability is fine, having a sensible default for the CMB source plane is important, since this value is effectively fixed in practice.

At this point, the core issue is that CMB lensing should not be modeled as an N(z) tracer at all, as it introduces conceptual inconsistencies and concrete technical failures in the connectors. I am currently unsure how to undo this cleanly within the existing structure, but I think reverting this design choice is necessary to have a robust CMB lensing implementation.

There are also problem with SACC as CMB Kappa is not considered a `NZTracer` in SACC (see example [here](https://github.com/LSSTDESC/sacc/blob/master/examples/CMB_LSS_write.ipynb): 

```python
# GC
s.add_tracer('NZ', 'gc',  # Name
             quantity='galaxy_density',  # Quantity
             spin=0,  # Spin
             z=z,  # z
             nz=nz)  # nz
# WL
s.add_tracer('NZ', 'wl',  # Name
             quantity='galaxy_shear',  # Quantity
             spin=2,  # Spin
             z=z,  # z
             nz=nz,  # nz
             extra_columns={'error': 0.1*nz},  # You can include extra columns for the N(z)
             sigma_g=0.28)  # You can add any extra information as **kwargs
# CMBK
s.add_tracer('Map', 'ck',  # Name
             quantity='cmb_convergence',  # Quantity
             spin=0,  # Spin
             ell=ell_beam, beam=beam)  # Beam
```

This complicates things also when using this SACC structure to call `TJPCov`. We had to adapt a pipeline that uses CMBK correctly in sacc when talking to TJPCov, but then later converting it to the wrong `NZTracer` type when interacting with firecrown.

**To Reproduce**
1. Try to run a CMB Lensing x Something case with cosmosis. 

**Expected behavior**
1. CMB Lensing runs within cosmosis and other connectors. 

**Observed behavior**
A clear and concise description of what you observed to happen.

**Output**
If necessary, add the SACC file, configuration files, error message, or any other output
that can help us understand the problem.

**Configuration:**
Master branch and since CMB Lensing has been merged into the master branch.

**Additional context**

vitenti:
## Description

This PR generalizes Firecrown's tracer system to support multiple types of projected fields beyond galaxy redshift distributions, with CMB lensing as the first new implementation. The refactoring introduces a flexible `ProjectedField` protocol that enables the framework to work with different types of tracers while maintaining backwards compatibility.

Fixes #606, #607 

## Type of change

* Refactoring

## Checklist:


marcpaterno (commented):
The only concern I have is with backward compatibility.  This PR takes care to support existing code that `imports` the old names.  But we may break user code that relies on existing serialized YAML that contains old class names. Do we need to support reading old serialized data?

I would like to be able to say that we explicitly do not, because I think trying to do so will substantially complicate the code for what I suspect is little value. But perhaps we need to get agreement in a Firecrown/CCL meeting that the value is small.

firecrown/app/analysis/_analysis_builder.py
arthurmloureiro, line 1: This part confuses me a little bit. How and where do we set this for CMB Lensing?

firecrown/metadata_functions/_extraction.py
```diff
@@ -56,6 +56,63 @@ def extract_all_tracers_inferred_galaxy_zdists(
...
+
+
+def extract_all_tracers_projected_fields(
+    sacc_data: sacc.Sacc,
+    allow_mixed_types: bool = False,
+) -> list[mdt.ProjectedField]:
+    """Extracts the two-point function metadata from a Sacc object.
+
+    Each SACC ``NZTracer`` is returned as a ``TomographicBin`` and each SACC
+    ``MapTracer`` is returned as a ``CMBLensing``. The ``z_lss`` value for a
```
arthurmloureiro, line 72: > each SACC ``MapTracer`` is returned as a ``CMBLensing``.

CMBLensing is a SACC ``MapTracer`` but not all MapTracers are CMBLensing, e.g. ISW.
I believe the next steps is adding ISW and SZ cross-correlations, so we just need to make sure we do not lock in CMBLensing being the only MapTracer without being easy to expand that :)

arthurmloureiro (commented):
There's a lot here I do no know about (cobaya connectors etc) but it seems to me that things are much easier to understand now. Projected fields and Tomographic Bins are way clearer :) 

Great job!!!

firecrown/metadata_functions/_extraction.py
vitenti: @arthurmloureiro thanks for the feedback! I updated the docstring to explain how we should handle future extensions.

arthurmloureiro (approved):

